PR #1, opened 2019-08-09
Add Marketplace apis

awgreene:
This commit introduces the OperatorSource and CatalogSourceConfig APIs
to the project.


estroz:
Does a global project structure make sense for the `api` repo, considering that we're adding several projects that have different structural requirements? For example, having a global `pkg` directory makes sense, but not `cmd/manager/main.go`.

Perhaps a better construction would be placing all APIs in `pkg/apis`, and all other project dirs/files in a `{project-name}` dir. It is possible to have both a [global and child `go.mod`'s](https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories) too, although not recommended.

njhale:
We may want to drop this if we're not going to follow through w/ moving our APIs to this repo.

(see: https://github.com/operator-framework/api/pull/3#issuecomment-546986038)

ecordell:
Plans have changed here - we won't be moving marketplace apis here after all.

PR #2, opened 2019-09-24
doc/proposals: proposal template and validator library proposal

estroz:
* `doc/proposals/TEMPLATE.md`: proposal template, a slimmed down version of the OKD [enhancement proposal template](https://github.com/derekwaynecarr/openshift-enhancements/blob/templates/enhancements/enhancements-template.md)
* `doc/proposals/validation-library.md`: discusses a possible implementation for an extentensible validation library.

Note: see the `Observations and open questions` section of `validation-library.md` for discussion points.

/cc @gallettilance @shawn-hurley @ecordell

### DIFF
--- a/doc/proposals/validator-library.md
+++ b/doc/proposals/validator-library.md
@@ -1,0 +1,201 @@
+# Title
+
+Status: `Draft`
+
+Version: `alpha`
+
+Implementation Owner: @estroz
+
+#### Table of Contents
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-goals](#non-goals)
+- [Proposal](#proposal)
+    - [Design Overview](#design-overview)
+    - [Implementation details](#implementation-details)
+    - [Risks and mitigations](#risks-and-mitigations)
+- [Observations and open questions](#observations-and-open-questions)
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


doc/proposals/validator-library.md
```diff
@@ -0,0 +1,201 @@
...
+type ValidatorFuncs []ValidatorFunc
+
+// Validator is an interface for validating arbitrary objects. A Validator
+// returns a set of functions that each validate some underlying object.
+// This allows the implementer to easily break down each validation step
+// into discrete units, and perform either lazy or active validation.
+type Validator interface {
+	// GetFuncs takes a list of arbitrary objects and returns a set of functions
+	// that each validate some object from the list.
+	GetFuncs(...interface{}) ValidatorFuncs
```
shawn-hurley: Would it be simpler to deal with one object at a time? (or async) such that for a given object we get the functions and run them, may make it easier if the `validateCRD` func conformed to the interface instead of wrapping everything in a closure? 

Would this allow for a simpler validate function code?

estroz: I agree. I've updated #3 to use @njhale's suggested simplified `Validator` interface. Once that has been agreed upon I'll update this proposal, or close this PR if we find it unnecessary.

PR #11, opened 2019-12-17
Have a Makefile to install operator-verify

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>


ecordell:
I think this is a great start! Thanks!
/lgtm

PR #12, opened 2020-01-17
Refactor validate CRD to fix conversion error

dinhxuanvu:
1. Fix conversion error due to incompatible type
1. Refactor validate CRD code to reduce duplication
2. Support v1.CRD type

### DIFF
--- a/pkg/validation/internal/crd_test.go
+++ b/pkg/validation/internal/crd_test.go
@@ -1,0 +1,82 @@
+package internal
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/operator-framework/api/pkg/validation/errors"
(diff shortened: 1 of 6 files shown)


pkg/validation/internal/crd_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	for _, tt := range table {
+		b, err := ioutil.ReadFile(tt.directory)
+		if err != nil {
+			t.Fatalf("Error reading CRD path %s: %v", tt.directory, err)
+		}
+
+		results := []errors.ManifestResult{}
+		switch tt.version {
+		case "v1":
```
jpeeler: I'd make this default rather than the beta version. But really only if you want to.

PR #19, opened 2020-02-28
GitHub action

Bowenislandsong:
This commit enables Github Action. The uses of external images need sha based tag whereas GitHub action repos can use normal version tag.


.github/workflows/go.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches: 
+      - master
+
+jobs:
+
```
awgreene: Please remove this empty line.

.github/workflows/go.yml
```diff
@@ -0,0 +1,27 @@
+name: Go
```
awgreene: Please update this file type to `.yaml`

.github/workflows/go.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches: 
+      - master
+
```
awgreene: Please remove this empty newline.

awgreene: Please do this in all places.

benluddy:
/lgtm

I agree with @awgreene that the extra newlines aren't helping readability for me. Are there any sets of published YAML style rules with significant adoption?

.github/workflows/go.yml
Bowenislandsong: lol, yeah, github propagated this format by default, so I went with it... lol. changing it,

.github/workflows/go.yml
Bowenislandsong: this file type is .yaml. This is the name of the ... action .... won't it be weird to have an action called go.yaml?

Bowenislandsong: wait ... this type is .... yml ......

Bowenislandsong:
> /lgtm
> 
> I agree with @awgreene that the extra newlines aren't helping readability for me. Are there any sets of published YAML style rules with significant adoption?

I mostly try to keep the default style that github provides. The main thing was to keep each test separate, but since you all don't like it,,, I changed it.

PR #22, opened 2020-03-10
chore(deps): Bump to kube 1.17

anik120:
Updates the kube libraries to the versions used in kube 1.17


go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	github.com/sirupsen/logrus v1.4.2
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.4.0
-	k8s.io/api v0.16.7
```
Bowenislandsong: should we use `replace` to pin the version? also, since we are following origin, should we use exactly the version they are using? Thanks.

anik120: @Bowenislandsong thanks for reviewing the PR!
I see that origin is using 1.17.1, thanks for pointing that out. But could you help me understand why we'd wanna keep pace with origin? I don't see any dependency between api and origin.
Also, I don't get the need to use `replace` in this case.

ecordell: Aligning with origin will help when we need to pull api in olm, which also depends on origin for ClusterOperator apis.

(that said I think this patch version will be fine - I'm not sure that it's a big deal to differ in patch versions as long as we're prepared to resolve any inconsistencies if we hit them)

anik120: @ecordell thanks for clarifying that. Is that an lgtm or a let's change the patch version anyway though? :smile:

PR #27, opened 2020-04-23
refactor(crds): decode to v1 directly w/o conversions

njhale:
Replace hub and spoke CRD conversion with direct unmarshaling to v1.
This follows [advice](https://github.com/operator-framework/operator-lifecycle-manager/pull/1470#discussion_r412975584) to avoid unversioned k8s apiserver types and conversions client-side.

It's also worth noting that using versioned CRD types will miss any new fields added until the module is updated. An alternative would be to decode and return `Unstructured`.

### DIFF
--- a/crds/defs.go
+++ b/crds/defs.go
@@ -8,11 +8,7 @@ import (
 	"fmt"
 	"sync"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/install"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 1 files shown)


crds/defs.go
```diff
@@ -37,18 +35,17 @@ func (c crdFile) mustUnmarshal() *apiextensionsv1.CustomResourceDefinition {
 	reader := bytes.NewReader(data)
 	decoder := yaml.NewYAMLOrJSONDecoder(reader, 30)
 	if err = decoder.Decode(u); err != nil {
```
benluddy, line 37: Is it necessary to unmarshal twice? As far as I can tell, it's safe to decode directly into the v1 struct and sanity check its TypeMeta before trusting its other fields.

njhale: That's a good point! Let's do that.

PR #31, opened 2020-04-30
Update OLM to use UID for OG Labels

awgreene:
Problem:
OLM applies an "OperatorGroup Label" to namespaces which makes it easy
to select namespaces that are included in an OperatorGroup.

Currently, OLM applies a label with a blank value whose key is equal to
"olm.operatorgroup/OperatorGroup Namespace.OperatorGroup Name".
Kubernetes limits the lengths of label values and keys to 63 characters.
This limit can easily be overcome when the OperatorGroup has a long name
or when it is deployed in a namespace with a long name.

Solution:
Update OLM to use "olm.operatorgroup.uid/<OperatorGroup UID>" as the key for
OperatorGroup labels. The length of this label will always be 58
characters as UIDs are 36 characters long.

### DIFF
--- a/pkg/operators/v1/operatorgroup_types.go
+++ b/pkg/operators/v1/operatorgroup_types.go
@@ -17,8 +17,8 @@ const (
 
 	OperatorGroupKind = "OperatorGroup"
 
-	OperatorGroupLabelPrefix   = "olm.operatorgroup/"
-	OperatorGroupLabelTemplate = OperatorGroupLabelPrefix + "%s.%s"
+	OperatorGroupLabelPrefix   = "olm.operatorgroup.uid/"
+	OperatorGroupLabelTemplate = OperatorGroupLabelPrefix + "%s"
 )
 
 // OperatorGroupSpec is the spec for an OperatorGroup resource.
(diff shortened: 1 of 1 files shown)


pkg/operators/v1/operatorgroup_types.go
```diff
@@ -108,16 +108,9 @@ func (o *OperatorGroup) HasServiceAccountSynced() bool {
...
-// GetLabel returns a label that is applied to Namespaces to signify that the
-// namespace is a part of the OperatorGroup using selectors.
-func (o *OperatorGroup) GetLabel() string {
-	key, _ := o.OGLabelKeyAndValue()
-	return key
-}
-
 // OGLabelKeyAndValue returns a key and value that should be applied to namespaces listed in the OperatorGroup.
 func (o *OperatorGroup) OGLabelKeyAndValue() (string, string) {
-	return fmt.Sprintf(OperatorGroupLabelTemplate, o.GetNamespace(), o.GetName()), ""
```
benluddy, line 120: Should this method return an error if `o.UID` is empty, to catch the possibility of invalid usage? I suppose even in the existing implementation, there's a risk that Name and Namespace are unset.

awgreene: I like the idea of improving the existing experience though! I'll update it!

awgreene: Updated the method to return an error.

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -108,35 +108,36 @@ func (o *OperatorGroup) HasServiceAccountSynced() bool {
...
-
 // OGLabelKeyAndValue returns a key and value that should be applied to namespaces listed in the OperatorGroup.
-func (o *OperatorGroup) OGLabelKeyAndValue() (string, string) {
-	return fmt.Sprintf(OperatorGroupLabelTemplate, o.GetNamespace(), o.GetName()), ""
+// If the UID is not set an error is returned.
+func (o *OperatorGroup) OGLabelKeyAndValue() (string, string, error) {
+	if string(o.UID) == "" {
+		return "", "", fmt.Errorf("Missing UID")
+	}
+	return fmt.Sprintf(OperatorGroupLabelTemplate, string(o.UID)), "", nil
```
Bowenislandsong, line 117: maybe use `GetUID()` so that we can get rid of the `error` return here. Since the previous implementation also considers not getting name/ns is not an error to be handled here. It is invalid for a server to return an object that does not have UID.
Wonder if that makes sense, else just a comment about using GetUID(). thanks!

Bowenislandsong: nvm, since it was @benluddy's idea to add it.

njhale: We should probably still use `GetUID()` since it's the appropriate getter.

awgreene: Updated.

PR #29, opened 2020-04-24
feat: Enable bundle validation with v1 CRD content

dinhxuanvu:
registry.Bundle now can contain either v1 or v1beta CRD and bundle
validation will account for 2 different types of CRD while checking
if CRD exists or not.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

PR #32, opened 2020-05-05
validation: bundle CRDs are compared by definition key instead of name

estroz:
CRDs should not be compared by name because each CSV `customresourcedefinition` element is not unique by name only, i.e. multiple versions of the same CRD are allowed.

Also fixed a test comparison (wrong argument order).

/kind bug

/cc @dinhxuanvu @kevinrizza

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -1,13 +1,11 @@
 package internal
 
 import (
-	"encoding/json"
 	"fmt"
 
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
 
-	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-registry/pkg/registry"
 )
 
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
Looking good but hold for now.
/hold

pkg/validation/internal/bundle.go
```diff
@@ -24,71 +22,82 @@ func validateBundles(objs ...interface{}) (results []errors.ManifestResult) {
...
 		return nil, errors.ErrInvalidParse("error getting bundle CRDs", err)
 	}
-	crdNames := map[string]struct{}{}
+	keySet := map[registry.DefinitionKey]struct{}{}
 	for _, crd := range crds {
-		crdNames[crd.GetName()] = struct{}{}
+		if len(crd.Spec.Versions) == 0 {
+			// Skip group, which CSVs do not support.
+			key := registry.DefinitionKey{
+				Name:    crd.GetName(),
```
exdx, line 78: Our work adding v1 crds to bundles is going to break this - instead of returning a `*v1beta1.CustomResourceDefinition` we are returning an interface now. We started addressing this in https://github.com/operator-framework/api/pull/29/files#diff-8a818e661938b6d25ef2013d1289f89cR94
but I believe we are planning to move this code back into the registry instead - right @dinhxuanvu? 

See https://github.com/operator-framework/operator-registry/pull/295 for the PR with the breaking change. we are planning to merge this one this week. 

We can merge as-is and then fix it in a follow-up, or hold off on this change.

dinhxuanvu: This PR is meant to fix api as it is. We will address the v1 CRD support in subsequent PR.

exdx: Ok, sounds good

PR #34, opened 2020-05-05
Update webhook description

awgreene:
Replace webhook description `Name` field with `GenerateName` to match new semantics.

for background, see https://github.com/operator-framework/operator-lifecycle-manager/pull/1489#discussion_r420444398


njhale:
Looks like were missing the deepcopy changes. Running `make generate` should update them.

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -124,7 +124,7 @@ type CRDDescription struct {
 // APIServiceDescription provides details to OLM about apis provided via aggregation
 // +k8s:openapi-gen=true
 type APIServiceDescription struct {
-	Name              string                 `json:"name"`
+	GenerateName      string                 `json:"generateName"`
```
njhale, line 127: Don't we want to change the field in `WebhookDescription`, not `APIServiceDescription`?

awgreene: Resolved.

njhale:
/approve

This looks good to me!

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -182,7 +183,7 @@ type WebhookDescription struct {
 // GetValidatingWebhook returns a ValidatingWebhook generated from the WebhookDescription
 func (w *WebhookDescription) GetValidatingWebhook(namespace string, namespaceSelector *metav1.LabelSelector, caBundle []byte) admissionregistrationv1.ValidatingWebhook {
 	return admissionregistrationv1.ValidatingWebhook{
-		Name:                    w.Name,
+		Name:                    w.GenerateName,
```
benluddy, line 186: Is this potentially problematic?

awgreene: Shouldn't be given that:
- OLM ensures that the GenerateName is unique
- The fact that ValidatingWebhook is not a cluster resource, ValidatingWebhookConfiguration is.

PR #35, opened 2020-05-13
Remove registry dependency

kevinrizza:


pkg/manifests/loader.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil
+	}
+
+	fileReader, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("unable to load file %s: %s", path, err)
+	}
+
+	decoder := yaml.NewYAMLOrJSONDecoder(fileReader, 30)
+	csv := unstructured.Unstructured{}
```
estroz, line 75: There could be more than one manifest per file.

pkg/validation/test_data/invalid_bundle/etcdbackups.etcd.database.coreos.com.crd.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
estroz, line 1: Can you rename `pkg/validation/test_data` to `pkg/validation/testdata`?

pkg/manifests/types.go
```diff
@@ -0,0 +1,37 @@
+package manifests
```
estroz, line 1: Perhaps this makes more sense in a `metadata` package?

pkg/manifests/loader.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		errs = append(errs, fmt.Errorf("error loading objs in directory: %s", err))
+	}
+
+	if bundle == nil || bundle.Csv == nil {
+		errs = append(errs, fmt.Errorf("no bundle csv found"))
+		return utilerrors.NewAggregate(errs)
+	}
+
+	/*
```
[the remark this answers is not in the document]
kevinrizza: I think it is still valid as a validation of the CSV itself to say "Hey you said you provided these APIs, is that actually the case in your bundle?"

My intention was actually that this check should be defined as part of the validation. It seems very weird to do this specifically for this subset when many of the other bundle validations are implemented in the validation package. It only happened initially that way, I assume, because this code was originally embedded into the registry serialization code.

Commenting it out was supposed to remind me to do that, but it seems like I just overlooked it. I'll update this to reimplement it.

pkg/manifests/loader.go
kevinrizza: While this is a totally fair point, and one we should address, currently the registry doesn't parse multi manifest yaml files either. IMO if we want to allow them we should update the registry first so that we aren't allowing things that won't work on the unpacking side.

pkg/manifests/directory.go
```diff
@@ -1,31 +1,15 @@
...
-	"github.com/operator-framework/api/pkg/validation"
-	"github.com/operator-framework/api/pkg/validation/errors"
-
-	"github.com/operator-framework/operator-registry/pkg/registry"
-)
-
-// GetManifestsDir parses all bundles and a package manifest from dir, which
-// are returned if found along with any errors or warnings encountered while
-// parsing/validating found manifests.
-func GetManifestsDir(dir string) (registry.PackageManifest, []*registry.Bundle, []errors.ManifestResult) {
```
estroz, line 16: This is still something the SDK needs for [`run packagemanifests`](https://github.com/operator-framework/operator-sdk/blob/master/internal/olm/operator/manager.go#L74). Will a function that returns a `PackageManifest` and `[]Bundle` exist somewhere else?

pkg/validation/validation.go
```diff
@@ -13,9 +13,6 @@ import (
 	"github.com/operator-framework/api/pkg/validation/internal"
 )
 
-// PackageManifestValidator implements Validator to validate package manifests.
-var PackageManifestValidator = internal.PackageManifestValidator
```
estroz, line 17: I assume this will be moved to operator-registry, like the bundle validator was? The SDK uses this validator but not the `PackageUpdateGraphValidator`.

pkg/manifests/bundleloader.go
```diff
@@ -0,0 +1,177 @@
...
+// are part of the bundle.
+func loadBundle(csvName string, dir string) (*Bundle, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	var errs []error
+	bundle := &Bundle{
+		Name: csvName,
```
estroz, line 111: Quick q: would this make more sense as a plain semver string?

pkg/manifests/bundleloader.go
```diff
@@ -0,0 +1,177 @@
...
+					&crd)
+				if err != nil {
+					return nil, fmt.Errorf("unable to parse CRD %s: %s", f.Name(), err.Error())
+				}
+				bundle.V1CRDs = append(bundle.V1CRDs, &crd)
+			} else {
+				return nil, fmt.Errorf("unsupported CRD version %s for %s", version, f.Name())
+			}
+		default:
+			bundle.Objects = append(bundle.Objects, obj)
```
estroz, line 172: One difference I found between this implementation and that in OR: all objects in a bundle, including CSVs and CRDs, are [appended](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/directory.go#L224) to `Bundle.Objects`. Is it reasonable to have the same behavior here?

pkg/manifests/bundleloader.go
kevinrizza: In the future, definitely, especially given the future where a csv is no longer required in future bundle versions. For backwards compatibility though, we need to use this as the unique identifier because version itself is an optional field.

pkg/manifests/bundleloader.go
kevinrizza: Ah, yes, it is totally reasonable and the correct behavior. I'll push a change to this pr

pkg/validation/internal/package_manifest.go
```diff
@@ -3,31 +3,32 @@ package internal
...
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
-
-	"github.com/operator-framework/operator-registry/pkg/registry"
 )
 
 var PackageManifestValidator interfaces.Validator = interfaces.ValidatorFunc(validatePackageManifests)
 
 func validatePackageManifests(objs ...interface{}) (results []errors.ManifestResult) {
+	fmt.Println("trying to validate package manifests")
```
estroz, line 14: @kevinrizza some leftover print statements

pkg/validation/internal/package_manifest.go
```diff
@@ -3,31 +3,32 @@ package internal
...
 	for _, obj := range objs {
 		switch v := obj.(type) {
-		case *registry.PackageManifest:
+		case *manifests.PackageManifest:
+			fmt.Println("found a package manifest in the list of objects")
```
estroz, line 18: Here too

pkg/validation/internal/package_manifest.go
kevinrizza: ouch, okay updated :)

pkg/manifests/bundleloader.go
```diff
@@ -0,0 +1,177 @@
...
+				&csv)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse CSV %s: %s", f.Name(), err.Error())
+			}
+			bundle.CSV = &csv
+		case "CustomResourceDefinition":
+			version := obj.GetAPIVersion()
+			if version == apiextensionsv1beta1.SchemeGroupVersion.String() {
+				crd := apiextensionsv1beta1.CustomResourceDefinition{}
+				err := runtime.DefaultUnstructuredConverter.FromUnstructured(
```
dinhxuanvu, line 154: This string-to-unstructured-to-crd conversion is the root cause of `int64 to float64` error. So we should avoid converting `unstructured` to a specific type. Should only use `unstructured` to determine the object type and then convert string data straight to the object type like CRD. See: https://github.com/operator-framework/operator-registry/pull/334/files#diff-b754381164690e8ba263f719136845d3R320

estroz: Is this behavior defined for unstructured conversion?

estroz: I think this issue can be addressed in a follow-up.

dinhxuanvu: I'm fine with fixing this at another time. Please make a note somewhere so it won't get lost.

PR #36, opened 2020-05-21
WebhookDesc should use containerPort field

awgreene:
Problem: The CSV  ignores the csv.Spec.WebhookDesc.ContainerPort field
when creating the validating / mutating webhook configuration.

Solution: Wire the port into the webhook configurations clientConfig.


awgreene:
Updated the testcases to use non-nil values.

PR #38, opened 2020-05-27
pkg/manifests: `loadBundle` returns an error if a subdir or hidden file is found

estroz:
A valid bundle manifests directory should not contain subdirectories or hidden files, and have helpful error messages when it does.

/cc @dinhxuanvu @kevinrizza @gallettilance


pkg/manifests/testdata/invalid_bundle_with_subdir/test.example.com_tests_crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1
```
gallettilance, line 1: can we test with v1beta since this was the cause of this change IIRC?

estroz:
Feel free to merge whenever

PR #39, opened 2020-06-02
validation: bundle CRDs are compared by definition key instead of name (#32)

estroz:
#32 was reverted at some point. This PR fixes the reversion, considering both v1 and v1beta1 CRDs.

/cc @dinhxuanvu @kevinrizza @gallettilance 

/kind bug

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -2,11 +2,13 @@ package internal
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/operator-framework/api/pkg/manifests"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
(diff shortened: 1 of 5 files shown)


pkg/validation/interfaces/validator.go
```diff
@@ -44,3 +46,18 @@ func (validators Validators) Validate(objs ...interface{}) (results []errors.Man
 func (validators Validators) WithValidators(vals ...Validator) Validators {
 	return append(vals, validators...)
 }
+
+// DefinitionKey is a unqiue identifier for an API definition.
+type DefinitionKey struct {
```
kevinrizza, line 51: Is this type strictly necessary? Can we just stringify the gvk and compare that directly, rather than adding this type definition just to include the `String()` function?

estroz: That would work although I don't like breaking the semantics of a GVK just for a string. Also I could see this being used elsewhere in the future, like its counterpart in [operator-registry](https://github.com/operator-framework/operator-registry/blob/v1.12.4/pkg/registry/types.go#L35).

estroz: Removed.

Issue #40, opened 2020-06-03
No License


magescher:
I was unable to find a license file associated with the code base. Is this intentional? If not, when/how will this be licensed?

dvattaka:
Hi, Is the operator-framework/api/ is licensed under [Apache License 2.0](https://github.com/operator-framework/api/blob/149a621385f724bac40499ff69f7feb98396151c/LICENSES/Apache-2.0.txt)? If that's the case, could you please add the LICENSE file?

benluddy:
Yes! Thank you for filing an issue. The license can now be found at https://github.com/operator-framework/api/blob/master/LICENSE.

PR #41, opened 2020-06-04
Add License

awgreene:


awgreene:
(Possibly) addresses https://github.com/operator-framework/api/issues/40

estroz:
This should probably be added to all go files.

Bowenislandsong:
> A license header should probably be added to all go files.

@awgreene I think you can use https://github.com/elastic/go-licenser to address this comment.

awgreene:
> @awgreene I think you can use https://github.com/elastic/go-licenser to address this comment.

That seems to work, have we used this tool in other repos? @njhale @ecordell is there Licensor that I should use given the planned contribution to CNCF?

Bowenislandsong:
are we dropping red hat from the license? 

Could we please drop the year? I was not a fan of removing/updating those license years.

.reuse/dep5
```diff
@@ -0,0 +1,10 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: The Operator Framework API
+Upstream-Contact: The Operator Framework Authors <agreene@redhat.com>
+Source: https://github.com/operator-framework/api
+
+# Sample paragraph, commented out:
+#
+# Files: src/*
```
awgreene, line 8: Looks like we could use this instead of the make file, thoughs?

dvattaka:
Hi, Is the operator-framework/api/ is Licensed under Apache License 2.0? If that's the case, could you please add the LICENSE file?

estroz:
@awgreene would be nice to get this in soon.

ankitathomas:
Closing this out, https://github.com/operator-framework/api/pull/57 add the Apache 2.0 License

PR #42, opened 2020-06-08
OperatorHub io validator

kevinrizza:
This pr attempts to create parity in the API project with the existing operator-courier tool's validation command which validates package manifest format operator bundles. It does this only for the new operator bundle image directory format. It updates and resolves several issues with existing validators, as well as introduces a new optional validator for OperatorHub.io validation.

Additionally, it overhauls the internal test `operator-verify` command line tool to serve as an example for import implementation.

### DIFF
--- a/pkg/manifests/bundle.go
+++ b/pkg/manifests/bundle.go
@@ -20,3 +20,17 @@ type Bundle struct {
 	V1CRDs         []*apiextensionsv1.CustomResourceDefinition
 	Dependencies   []*Dependency
 }
+
+func (b *Bundle) ObjectsToValidate() []interface{} {
+	objs := []interface{}{}
+	for _, crd := range b.V1CRDs {
+		objs = append(objs, crd)
+	}
+	for _, crd := range b.V1beta1CRDs {
+		objs = append(objs, crd)
+	}
+	objs = append(objs, b.CSV)
+	objs = append(objs, b)
+
+	return objs
+}
--- a/pkg/manifests/bundleloader.go
+++ b/pkg/manifests/bundleloader.go
@@ -18,8 +18,9 @@ import (
 
 // bundleLoader loads a bundle directory from disk
 type bundleLoader struct {
-	dir    string
-	bundle *Bundle
+	dir      string
+	bundle   *Bundle
+	foundCSV bool
 }
 
 func NewBundleLoader(dir string) bundleLoader {
@@ -32,6 +33,12 @@ func (b *bundleLoader) LoadBundle() error {
 	errs := make([]error, 0)
 	if err := filepath.Walk(b.dir, collectWalkErrs(b.LoadBundleWalkFunc, &errs)); err != nil {
 		errs = append(errs, err)
+	}
+
+	if !b.foundCSV {
+		errs = append(errs, fmt.Errorf("unable to find a csv in bundle directory %s", b.dir))
+	} else if b.bundle == nil {
+		errs = append(errs, fmt.Errorf("unable to load bundle from directory %s", b.dir))
 	}
 
 	return utilerrors.NewAggregate(errs)
@@ -81,6 +88,8 @@ func (b *bundleLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err error)
 	if csv.GetKind() != operatorsv1alpha1.ClusterServiceVersionKind {
 		return nil
 	}
+
+	b.foundCSV = true
 
 	var errs []error
 	bundle, err := loadBundle(csv.GetName(), filepath.Dir(path))
--- a/pkg/validation/internal/csv.go
+++ b/pkg/validation/internal/csv.go
@@ -46,7 +46,7 @@ func validateCSV(csv *v1alpha1.ClusterServiceVersion) errors.ManifestResult {
 	// validate installModes
 	result.Add(validateInstallModes(csv)...)
 	// check missing optional/mandatory fields.
-	result.Add(checkFields(csv)...)
+	result.Add(checkFields(*csv)...)
 	return result
 }
 
@@ -67,7 +67,7 @@ func parseCSVNameFormat(name string) (string, semver.Version, error) {
 }
 
 // checkFields runs checkEmptyFields and returns its errors.
-func checkFields(csv *v1alpha1.ClusterServiceVersion) (errs []errors.Error) {
+func checkFields(csv v1alpha1.ClusterServiceVersion) (errs []errors.Error) {
 	result := errors.ManifestResult{}
 	checkEmptyFields(&result, reflect.ValueOf(csv), "")
 	return append(result.Errors, result.Warnings...)
--- a/pkg/validation/internal/operatorhub.go
+++ b/pkg/validation/internal/operatorhub.go
@@ -1,0 +1,152 @@
+package internal
+
+import (
+	"fmt"
+	"net/mail"
+	"net/url"
+	"strings"
+
+	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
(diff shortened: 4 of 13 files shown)


pkg/manifests/bundle.go
```diff
@@ -20,3 +20,17 @@ type Bundle struct {
 	V1CRDs         []*apiextensionsv1.CustomResourceDefinition
 	Dependencies   []*Dependency
 }
+
+func (b *Bundle) ObjsToValidate() []interface{} {
```
dinhxuanvu, line 24: I would prefer the full name here for func method `ObjectsToValidate` for clarity.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+	// Expect either `alm-examples` or `olm.examples` but not both
+	var examplesString string
+	almExamples, almOK := annotations["alm-examples"]
+	olmExamples, olmOK := annotations["olm.examples"]
+	if almOK {
+		if olmOK {
+			errs = append(errs, fmt.Errorf("both `csv.metadata.annotations.alm-examples` and `csv.metadata.annotations.olm.examples` are present"))
+		}
+		examplesString = almExamples
+	} else {
```
dinhxuanvu, line 170: I think this if condition is missing a case when alm and olm are both not present.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+	// parse alm examples
+	examples := []unstructured.Unstructured{}
+	dec := yaml.NewYAMLOrJSONDecoder(strings.NewReader(examplesString), 8)
+	if err := dec.Decode(&examples); err != nil && err != io.EOF {
+		errs = append(errs, errors.ErrInvalidParse("error decoding example CustomResource", err))
+		return errs
+	}
+	parsed := map[string]struct{}{}
+	for _, example := range examples {
+		parsed[example.GetObjectKind().GroupVersionKind().Kind] = struct{}{}
```
dinhxuanvu, line 183: What happens if we have multiple items with a same `Kind`? Is that case possible because if it is, then key collision will happen here.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// ensure there is a valid example for each owned crd
+	for _, crd := range csv.Spec.CustomResourceDefinitions.Owned {
+		if _, ok := parsed[crd.Kind]; !ok {
```
dinhxuanvu, line 188: This check is not sufficient as it only checks all CRDs exist in examples but not check if there is extra CR that doesn't exist in CRD list.

pkg/manifests/bundleloader.go
```diff
@@ -34,6 +35,12 @@ func (b *bundleLoader) LoadBundle() error {
 		errs = append(errs, err)
 	}
 
+	if b.foundCSV == false {
```
estroz, line 38: Nit
```suggestion
	if !b.foundCSV {
```

pkg/validation/internal/csv.go
```diff
@@ -67,7 +67,7 @@ func parseCSVNameFormat(name string) (string, semver.Version, error) {
 }
 
 // checkFields runs checkEmptyFields and returns its errors.
-func checkFields(csv *v1alpha1.ClusterServiceVersion) (errs []errors.Error) {
+func checkFields(csv v1alpha1.ClusterServiceVersion) (errs []errors.Error) {
```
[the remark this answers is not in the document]
estroz: Is having a pointer parameter a pain point?

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
+	result := errors.ManifestResult{Name: bundle.Name}
+
+	if bundle == nil {
+		result.Add(errors.ErrInvalidBundle("Bundle is nil", nil))
+		return result
+	}
+
+	if bundle.CSV == nil {
+		result.Add(errors.ErrInvalidBundle("Bundle csv is nil", bundle.Name))
```
estroz, line 40:
```suggestion
		result.Add(errors.ErrInvalidBundle("Bundle csv is nil", bundle.Name))
		return result
```
The pointer indirection below will panic otherwise

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+		if link.URL != "" {
+			_, err := url.ParseRequestURI(link.URL)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("csv.Spec.Links url %s invalid", link.URL))
+			}
+		}
+	}
+
+	if capability, ok := csv.ObjectMeta.Annotations["capabilities"]; ok {
+		validCapabilities := map[string]struct{}{
```
estroz, line 85: This would be clearer as a global variable.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	if csv.Spec.Icon != nil {
+		// only one icon is allowed
+		if len(csv.Spec.Icon) != 1 {
+			errs = append(errs, fmt.Errorf("csv.Spec.Icon should only have one element"))
+		}
+
+		for _, icon := range csv.Spec.Icon {
```
estroz, line 104: Should all icons be validated if there should be exactly one? Was this intended to be a no-op if `len(csv.Spec.Icon) == 0`?

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+	} else {
+		errs = append(errs, fmt.Errorf("csv.Spec.Icon not specified"))
+	}
+
+	if categories, ok := csv.ObjectMeta.Annotations["categories"]; ok {
+		categorySlice := strings.Split(categories, ",")
+
+		validCategories := map[string]struct{}{
```
estroz, line 129: Global variable.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,194 @@
...
+			if _, ok := validCategories[category]; !ok {
+				errs = append(errs, fmt.Errorf("csv.Metadata.Annotations.Categories %s is not a valid category", category))
+			}
+		}
+	}
+
+	return errs
+}
+
+func checkAlmExamples(csv v1alpha1.ClusterServiceVersion) []error {
```
estroz, line 156: I think this check is already done in the CSV validator.

anik120:
@kevinrizza 
From your PR commit message 
> It updates and resolves several issues with existing validators

Is is possible to enumerate the issues this PR attempts to solve for better summary?

kevinrizza:
> @kevinrizza
> From your PR commit message
> 
> > It updates and resolves several issues with existing validators
> 
> Is is possible to enumerate the issues this PR attempts to solve for better summary?

@anik120 They are enumerated in the commit messages:
> (fix) Actually validate csv fields
> > Pass by value to validator instead of pointer which ignores type
and
> *Expose new bundle function to get all objects for default validators

Small enough changes that I didn't want to clutter the actual purpose of the pr description, but were needed in order to actually create parity.

kevinrizza:
Addressed everyone's feedback as much as possible, please take another look.

pkg/validation/internal/operatorhub.go
```diff
@@ -0,0 +1,148 @@
...
+			errs = append(errs, fmt.Errorf("csv.Spec.Links elements should contain both name and url"))
+		}
+		if link.URL != "" {
+			_, err := url.ParseRequestURI(link.URL)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("csv.Spec.Links url %s invalid", link.URL))
+			}
+		}
+	}
+
```
estroz, line 110: So no panics occur but all relevant errors are returned:
```suggestion
	if csv.GetAnnotations() == nil {
		csv.SetAnnotations(make(map[string]string))
	}

```

kevinrizza:
@estroz okay, updated once again based on feedback. thanks!

PR #44, opened 2020-06-22
fix(crd): Fix CRD validation falsely error on unused CRD versions

dinhxuanvu:
CRD can have multiple versions and CSV can only reference one version.
CRD validation falsely errors out when there are unused versions in
provided CRDs.

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -43,13 +43,25 @@ func validateOwnedCRDs(bundle *manifests.Bundle, csv *operatorsv1alpha1.ClusterS
 	}
 
 	// All owned keys must match a CRD in bundle.
+	ownedGVSet := make(map[schema.GroupKind]struct{})
 	for _, ownedKey := range ownedKeys {
 		if _, ok := keySet[ownedKey]; !ok {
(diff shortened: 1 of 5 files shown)


estroz:
@dinhxuanvu will OLM ever install a non-stored version of an owned CRD? Can the validator only check `stored: true` CRD versions?

dinhxuanvu:
> @dinhxuanvu will OLM ever install a non-stored version of an owned CRD? Can the validator only check `stored: true` CRD versions?

I'm not sure what you mean by installing a version within a CRD. OLM installs the entire CRD and it is up to operator to use which version of CRD. The `owned` CRD section in CSV is to ensure we know which version in the CRD that this operator intends to use/control. There can be multiple served versions that can be used by multiple different operators.

dinhxuanvu:
@estroz Can I have a lgtm from you so I can merge this PR?

dinhxuanvu:
> This works although it would be simpler to reverse the order of comparison to check if a CRD version is an owned CRD instead of if an owned CRD has a corresponding CRD version. Something for a follow-up.
> 
> /lgtm

We still need to alert errors when a CRD is included in bundle and not present in CSV and vice versa so I don't think the order matters here.

PR #45, opened 2020-06-25
Approval: Clarify possible values

alvaroaleman:
Currently this field does not seem to get defaulted (should it?) and the docs are pretty useless because the name is self-explanatory for the what, but the how is not documented:

```
$ k explain subscription.spec.installPlanApproval
KIND:     Subscription
VERSION:  operators.coreos.com/v1alpha1

FIELD:    installPlanApproval <string>

DESCRIPTION:
     Approval is the user approval policy for an InstallPlan.
```

### DIFF
--- a/crds/operators.coreos.com_installplans.yaml
+++ b/crds/operators.coreos.com_installplans.yaml
@@ -62,6 +62,7 @@ spec:
             properties:
               approval:
                 description: Approval is the user approval policy for an InstallPlan.
+                  It must be one of "Automatic" or "Manual".
                 type: string
               approved:
                 type: boolean
--- a/crds/operators.coreos.com_subscriptions.yaml
+++ b/crds/operators.coreos.com_subscriptions.yaml
@@ -1634,6 +1634,7 @@ spec:
                               type: string
               installPlanApproval:
                 description: Approval is the user approval policy for an InstallPlan.
+                  It must be one of "Automatic" or "Manual".
                 type: string
               name:
                 type: string
--- a/pkg/operators/v1alpha1/installplan_types.go
+++ b/pkg/operators/v1alpha1/installplan_types.go
@@ -14,6 +14,7 @@ const (
 )
 
 // Approval is the user approval policy for an InstallPlan.
+// It must be one of "Automatic" or "Manual".
 type Approval string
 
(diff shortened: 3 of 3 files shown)


alvaroaleman:
@bipuladh @exdx @dinhxuanvu can one of you please review this? Its a tiny change that helps a lot with usability.

Also, it would be awesome if this repo would get an OWNERS file so it becomes clear who to assign changes to.

crds/zz_defs.go
```diff
@@ -98,7 +98,7 @@ func operatorsCoreosCom_catalogsourcesYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "operators.coreos.com_catalogsources.yaml", size: 7231, mode: os.FileMode(420), modTime: time.Unix(1591650274, 0)}
+	info := bindataFileInfo{name: "operators.coreos.com_catalogsources.yaml", size: 7231, mode: os.FileMode(436), modTime: time.Unix(1593116031, 0)}
```
exdx, line 101: this changed as a result of running codegen? A little confused as to why file permissions would change as a result of adding a comment to the source code.

alvaroaleman: yeah, as a result of `make generate`. I don't really understand this, I would not have expected this change but instead a change of the generated openapi schemes in `crds/operators.coreos.com_installplans.yaml` and `crds/operators.coreos.com_subscriptions.yaml`. To me, it looks like either I am using this somehow wrong or its broken. I was hoping there would be a job that fails and tells me what to do but doesn't seem to be the case :upside_down_face:

alvaroaleman:
@exdx can you advise on who to ping to get this merged?

crds/zz_defs.go
dinhxuanvu: This is just a bit strange. Maybe you can discard these changes from the PR and just keep the main change in `pkg/operators/installplan_types.go`. Then I can lgtm and merge this PR. Thanks.

alvaroaleman: @dinhxuanvu updated, PTAL

alvaroaleman:
@dinhxuanvu updated, ptal

PR #49, opened 2020-07-20
Add .spec.priority field on catalog source for dependency resolution

harishsurf:
This change allows cluster admins to define a
weight for their catalog so that it can be consumed by the
dependency resolver

Signed-off-by: Harish <hgovinda@redhat.com>


crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -90,6 +91,11 @@ spec:
                   a registry-server with. Only used when SourceType = SourceTypeGrpc.
                   If present, the address field is ignored.
                 type: string
+              priority:
+                description: Priority represents weighting catalogs for the dependency
```
Bowenislandsong, line 95: Reword this. Priority field, if specified, is used to rank catalog sources that contains operators to supply dependent APIs and prioritize certain catalogs based on the ranking.

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -58,6 +58,7 @@ spec:
           spec:
             type: object
             required:
+            - priority
```
Bowenislandsong, line 61: This is not a required field since we are trying to make it backwards compatible. Catalog sources without priority specified will be assigned a value.

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -90,6 +91,11 @@ spec:
                   a registry-server with. Only used when SourceType = SourceTypeGrpc.
                   If present, the address field is ignored.
                 type: string
+              priority:
+                description: Priority represents weighting catalogs for the dependency
+                  resolver
+                type: integer
+                format: int32
```
Bowenislandsong, line 98: Should we use int32? Is int ok? is there other fields also using int 32? Wdyt?

harishsurf: We could stick to `int`. [Similar `weight` usage ](https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/design-proposals/scheduling/podaffinity.md#api)

Bowenislandsong:
/lgtm
This PR should be followed by a version bump so that OLM and Registry and import it and use/generate artifacts.
/hold 
for a second review.

crds/operators.coreos.com_catalogsources.yaml
benluddy: The important thing here is that the range of acceptable values is precisely defined. If this field has type `integer`, and there's no additional information about which integers are acceptable, then the API is ambiguous.

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -90,6 +90,11 @@ spec:
                   a registry-server with. Only used when SourceType = SourceTypeGrpc.
                   If present, the address field is ignored.
                 type: string
+              priority:
+                description: Priority field, if specified, is used to rank catalog
```
benluddy, line 94: Is a catalog source with priority 1 preferred to a catalog source with priority 2? Or -1?

Bowenislandsong: right, we are defining that in the OLM logic. Here is just a type. Personally don't think we need to call it out here right?

crds/operators.coreos.com_catalogsources.yaml
Bowenislandsong: i'm ok either way, just want to compare to what is currently done.

crds/operators.coreos.com_catalogsources.yaml
njhale: As a consumer of an API, I would like to know what effect changes to this field will have. You may also want to check out the [CRD validation markers](https://book.kubebuilder.io/reference/markers/crd-validation.html), to see how we can  bound this fields and avoid headaches parsing bad values ourselves.

pkg/operators/catalogsource_types.go
```diff
@@ -29,6 +29,11 @@ type CatalogSourceSpec struct {
 	// SourceType is the type of source
 	SourceType SourceType
 
+	// Priority field, if specified, is used to rank catalog sources that contain operators
+	// to supply dependent APIs and prioritize certain catalogs based on the ranking.
```
njhale, line 33:
```suggestion
	// to supply dependencies and prioritize certain catalogs based on the ranking.
```

crds/operators.coreos.com_catalogsources.yaml
benluddy: The semantics should be unambiguous here. The API shouldn't define behaviors of its consumers, but it must clearly define both structure AND meaning. Otherwise, one consumer can decide that lower integers rank first, and another can decide that higher integers rank first, and neither one is wrong, because as far as the API is concerned, it's only a number divorced from meaning.

crds/operators.coreos.com_catalogsources.yaml
Bowenislandsong: Actually, according to https://book.kubebuilder.io/reference/markers/crd-validation.html, it seems like we can only us int for validation reasons?

benluddy: In the reference Harish found, it's not ambiguous because the API states the range of allowable integers precisely.

Bowenislandsong: right, right, but jumped thread here, i'm more saying that we should use `int`

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -90,6 +90,11 @@ spec:
...
+                  sources that contain operators to supply dependent APIs and prioritize
+                  certain catalogs based on the ranking.
+                type: integer
```
exdx, line 97: This sentence is very hard to understand. Could we try something simpler like

`Description: The priority assigns a weight to the catalog source that is then used during dependency resolution. Higher priorities indicate that this catalog source should be considered earlier in the dependency search.`

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -42,6 +42,11 @@ type CatalogSourceSpec struct {
 	// SourceType is the type of source
 	SourceType SourceType `json:"sourceType"`
 
+	// Priority field, if specified, is used to rank catalog sources that contain operators
```
exdx, line 45: Having "if specified" here is redundant since the parameter is marked as `+Optional`

crds/operators.coreos.com_catalogsources.yaml
harishsurf: I have rephrased it

crds/operators.coreos.com_catalogsources.yaml
harishsurf: The range would be bounded using [CRD validation markers](https://book.kubebuilder.io/reference/markers/crd-validation.html)

crds/operators.coreos.com_catalogsources.yaml
Bowenislandsong: Might want to include default priorities, default values, range, and behaviors if priorities are the same. basically a brief paraphrase to 

- The higher the value, the higher the priority. 
- The default priorities for Red Hat, Certified, and Community catalog sources are 10000,5000,1000  respectively. 
- The range of the priority value can go from positive to negative in the range of int32. 
- The default value to a catalog source with unassigned priority would be 0. This means, if cluster admins would like to make sure certain catalogs get considered last, they can assign the catalog with negative priority values. 
- The catalog source with the same priority values will be ranked lexicographically. 
-

pkg/operators/catalogsource_types.go
```diff
@@ -29,6 +29,11 @@ type CatalogSourceSpec struct {
...
 	SourceType SourceType
 
+	// Priority field assigns a weight to the catalog source to prioritize them so that it can be consumed by the dependency resolver.
+	// Higher weight indicates that this catalog source is preferred over lower weighted catalog sources during dependency resolution.
+	// +Optional
+	Priority int `json:"priority,omitempty"`
+
 	// ConfigMap is the name of the ConfigMap to be used to back a configmap-server registry.
 	// Only used when SourceType = SourceTypeConfigmap or SourceTypeInternal.
 	// +Optional
```
kevinrizza, line 39: Slightly derivative but, is this file the "latest" version of this type? Should we be manually modifying it at all? Or should there be some sync target in the makefile?

crds/operators.coreos.com_catalogsources.yaml
benluddy: >The default priorities for Red Hat, Certified, and Community catalog sources are 10000,5000,1000 respectively.

I'd omit this one. Creation of specific catalog sources is a usage of this API, not a semantic of the API itself.

kevinrizza: +1, that documentation should just live in the downstream component that builds these objects

pkg/operators/catalogsource_types.go
harishsurf: I will have a follow-up PR to remove these files as discussed in [slack](https://coreos.slack.com/archives/G3T7N42NP/p1595340592091300)

crds/operators.coreos.com_catalogsources.yaml
harishsurf: Updated the description

benluddy: The mentions of provided catalogs don't belong here -- provided catalogs are just catalogs. As Kevin said, that belongs in the documentation of the component that creates those catalogs.

pkg/operators/catalogsource_types.go
Bowenislandsong: might as well delete these from this PR.

PR #51, opened 2020-07-21
Operatorhub require csv version

kevinrizza:


awgreene:
Question: Should we be concerned that we might run into the issue where `version` is used instead of `versions`?

dinhxuanvu:
> Question: Should we be concerned that we might run into the issue where `version` is used instead of `versions`?

This is CSV version, not CRD. So there is no `versions` field.

PR #55, opened 2020-07-29
pkg/apis: add scorecard v1alpha3 API

estroz:
This PR moves [`pkg/apis/scorecard/v1alpha3`](https://github.com/operator-framework/operator-sdk/tree/v1.0.0-alpha.1/pkg/apis/scorecard/v1alpha3) from operator-sdk to a central location.

/cc @joelanford @shawn-hurley @ecordell @kevinrizza


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/google/go-cmp v0.4.0 // indirect
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
+	github.com/mattn/go-isatty v0.0.8
```
joelanford, line 12: I don't _love_ this dependency in the API repo. Would anyone complain too hard if we didn't colorize the output, ever?

pkg/apis/scorecard/v1alpha3/register.go
```diff
@@ -0,0 +1,14 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// SchemeGroupVersion is group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: "scorecard.operatorframework.io", Version: "v1alpha3"}
+
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
+	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
```
joelanford, line 13: Since we never need a client to be aware of these types, I think we can probably remove the `SchemeBuilder`. WDYT?

pkg/apis/scorecard/v1alpha3/register.go
```diff
@@ -0,0 +1,14 @@
+package v1alpha3
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// SchemeGroupVersion is group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: "scorecard.operatorframework.io", Version: "v1alpha3"}
```
joelanford, line 10: If we're removing `SchemeBuilder` and really have no need to integrate with schemes, should be rename this to simply `GroupVersion`?

go.mod
estroz: I'm ok with either removing tty checks completely, or changing the marshal function signature to `MarshalText(...MarshalOption)` where `WithColor` is an option (no color by default). Callers can import the tty package.

estroz: Regardless we can deal with this in a follow-up.

estroz:
@kevinrizza following typical API package layout, shouldn’t OLM’s APIs be under `pkg/apis/operators`? To me it makes sense to group all APIs under `pkg/apis` in subpackages named for their group.

PR #57, opened 2020-08-05
LICENSE: add apache 2.0 license

joelanford:
Add LICENSE file so our docs show up on `pkg.go.dev`.

Here's the license policy: https://pkg.go.dev/license-policy
Here's an example of what users see now: https://pkg.go.dev/github.com/operator-framework/api/pkg/operators/v1alpha1?tab=overview


dinhxuanvu:
I will close PR as there is already a PR for this #41

PR #59, opened 2020-08-06
Bug 1861605: Reduce length of explanation of status of Operator resource

simrandhaliw:
This PR rewords the explanation of status of Operator resource to reduce the character count.


crds/operators.coreos.com_operators.yaml
```diff
@@ -38,8 +38,8 @@ spec:
             description: OperatorSpec defines the desired state of Operator
             type: object
           status:
-            description: OperatorStatus describes the observed state of an operator
-              and its components.
+            description: OperatorStatus defines the observed state of Operator and
```
ecordell, line 41:
```suggestion
            description: OperatorStatus defines the observed state of an Operator and
```

PR #60, opened 2020-08-12
pkg/metadata: bundle metadata utils and types

estroz:
There should be a canonical set of utilities and types that can find and parse an `annotations.yaml` in a bundle directory instead of re-implementing that logic in various ways in various operator-framework repos. This repo is a good home for them because it is the source of truth for many OF APIs.

/cc @kevinrizza @dinhxuanvu 

/kind feature


estroz:
Relevant to this PR: because we're adding this function to the api lib, can we consider adding the bundle spec itself to this lib, specifically [labels](https://github.com/operator-framework/operator-registry/blob/v1.12.6/pkg/lib/bundle/generate.go#L27-L32) and some convenient access methods for them?

pkg/metadata/annotations.go
```diff
@@ -0,0 +1,82 @@
...
+func readAnnotations(fs afero.Fs, annotationsPath string) (map[string]string, error) {
+	// The annotations file is well-defined.
+	b, err := afero.ReadFile(fs, annotationsPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use the arbitrarily-indexed representation of the annotations file for forwards and backwards compatibility.
+	annotations := struct {
+		Annotations map[string]string `json:"annotations"`
```
kevinrizza, line 73: Why don't we use a more versioned structured type here? Just so that additional arbitrary annotations can be added? It seems like either way we need to validate that it has the set of required annotations to make the bundle work, so maybe that should be included here?

pkg/metadata/annotations.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, "", fmt.Errorf("metadata not found in %s", bundleRoot)
+	}
+
+	return annotations, annotationsPath, nil
+}
+
+// readAnnotations reads annotations from file(s) in bundleRoot and returns them as a map.
+func readAnnotations(fs afero.Fs, annotationsPath string) (map[string]string, error) {
+	// The annotations file is well-defined.
+	b, err := afero.ReadFile(fs, annotationsPath)
```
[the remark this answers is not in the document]
estroz: It's much easier to test with afero since we don't have to write testdata to disk for each new case. Plus there's a [Go design draft](https://go.googlesource.com/proposal/+/master/design/draft-iofs.md) for a std file system interface, which we could replace this with in the future.

pkg/metadata/annotations.go
estroz: Exactly, so arbitrary annotations can be added. There may be a way to load both versioned and arbitrary annotations into some struct type, which I can experiment with in this PR, but I wasn't sure about doing that initially (see https://github.com/operator-framework/api/pull/60#issuecomment-673081741).

pkg/metadata/directory.go
```diff
@@ -0,0 +1,76 @@
+package metadata
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
```
njhale, line 10: nice find!

pkg/metadata/annotations.go
estroz: Done

pkg/metadata/types.go
```diff
@@ -0,0 +1,130 @@
...
+	DefaultChannelName string `json:"operators.operatorframework.io.bundle.channel.default.v1" yaml:"operators.operatorframework.io.bundle.channel.default.v1"`
+
+	// AnnotationsRaw contains all other annotations in an annotations file
+	// that do not have keys that match the above tags.
+	AnnotationsRaw map[string]string `json:"-" yaml:"-"`
+}
+
+func (a AnnotationsV1) GetLabels() []string {
+	return []string{
+		"operators.operatorframework.io.bundle.mediatype.v1",
```
dinhxuanvu, line 107: Do we have these labels as constants somewhere so we can reuse them instead of hardcoding?

estroz: I can add these as constants.

bcrochet:
Is this still alive? I could sure use it. I'd be happy to rebase it and get it updated.

dinhxuanvu:
> Is this still alive? I could sure use it. I'd be happy to rebase it and get it updated.

Sure. Go ahead.

perdasilva:
closing as stale

PR #62, opened 2020-09-09
feat: bundle object validation

exdx:
Adds new validator type for object level validation in the bundle.

### DIFF
--- a/pkg/validation/internal/object.go
+++ b/pkg/validation/internal/object.go
@@ -1,0 +1,197 @@
+package internal
+
+import (
+	"encoding/json"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var ObjectValidator interfaces.Validator = interfaces.ValidatorFunc(validateObjects)
+
+const (
+	PodDisruptionBudgetKind     = "PodDisruptionBudget"
+	PriorityClassKind           = "PriorityClass"
+	RoleKind                    = "Role"
+	ClusterRoleKind             = "ClusterRole"
+	PodDisruptionBudgetAPIGroup = "policy"
+	SCCAPIGroup                 = "security.openshift.io"
+)
+
+// defaultSCCs is a map of the default Security Context Constraints present as of OpenShift 4.5.
+// See https://docs.openshift.com/container-platform/4.5/authentication/managing-security-context-constraints.html#security-context-constraints-about_configuring-internal-oauth
+var defaultSCCs = map[string]struct{}{
+	"privileged":       {},
+	"restricted":       {},
+	"anyuid":           {},
+	"hostaccess":       {},
+	"hostmount-anyuid": {},
+	"hostnetwork":      {},
+	"node-exporter":    {},
+	"nonroot":          {},
+}
+
+func validateObjects(objs ...interface{}) (results []errors.ManifestResult) {
+	for _, obj := range objs {
+		switch u := obj.(type) {
+		case *unstructured.Unstructured:
+			switch u.GroupVersionKind().Kind {
+			case PodDisruptionBudgetKind:
+				results = append(results, validatePDB(u))
+			case PriorityClassKind:
+				results = append(results, validatePriorityClass(u))
+			case RoleKind:
+				results = append(results, validateRBAC(u))
+			case ClusterRoleKind:
+				results = append(results, validateRBAC(u))
+			}
+		}
+	}
+	return results
+}
+
+// validatePDB checks the PDB to ensure the minimum and maximum budgets are set to reasonable levels.
+// See https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/adding-pod-disruption-budgets.md#limitations-on-pod-disruption-budgets
+func validatePDB(u *unstructured.Unstructured) (result errors.ManifestResult) {
+	pdb := policyv1beta1.PodDisruptionBudget{}
+
+	b, err := u.MarshalJSON()
+	if err != nil {
+		result.Add(errors.ErrInvalidParse("error converting unstructured", err))
+		return
+	}
+
+	err = json.Unmarshal(b, &pdb)
+	if err != nil {
+		result.Add(errors.ErrInvalidParse("error unmarshaling poddisruptionbudget", err))
+		return
+	}
+
+	/*
+	   maxUnavailable field cannot be set to 0 or 0%.
+	   minAvailable field cannot be set to 100%.
+	*/
+
+	maxUnavailable := pdb.Spec.MaxUnavailable
+	if maxUnavailable != nil && (maxUnavailable.IntVal == 0 || maxUnavailable.StrVal == "0%") {
+		result.Add(errors.ErrInvalidObject(pdb, "maxUnavailable field cannot be set to 0 or 0%"))
+	}
+
+	minAvailable := pdb.Spec.MinAvailable
+	if minAvailable != nil && minAvailable.StrVal == "100%" {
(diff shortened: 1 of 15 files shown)


pkg/validation/internal/object.go
```diff
@@ -0,0 +1,198 @@
...
+		if _, ok := set[item]; ok {
+			return true
+		}
+	}
+
+	return false
+}
+
+func containsDefaults(slice []string, defaults map[string]struct{}) bool {
+	for key := range defaults {
```
benluddy, line 192: Would it be better to flip this around so that it iterates over `slice` once time and checks each element against the map? I assume the inputs are always really small to this, but the growth rate is still bigger than it needs to be.

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+			result.Add(errors.ErrInvalidObject(pdb, "maxUnavailable field cannot be set to 0 or 0%"))
+		}
+	}
+
+	minAvailable := pdb.Spec.MinAvailable
+	if minAvailable != nil {
+		if minAvailable.StrVal == "100%" {
+			result.Add(errors.ErrInvalidObject(pdb, "minAvailable field cannot be set to 100%"))
+		}
+	}
```
ecordell, line 89:
```suggestion
	if minAvailable != nil && minAvailable.StrVal == "100%" {
		result.Add(errors.ErrInvalidObject(pdb, "minAvailable field cannot be set to 100%"))
	}
```

pkg/validation/internal/testdata/objects/invalid_role_create_pdb.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: default
+  name: pdb-modifier
+rules:
+  - apiGroups: ["policy"]
+    resources: ["poddisruptionbudgets"]
+    verbs: ["create", "list"]
```
anik120, line 9: nit: missing new line at the end of file.

Similar comments about the other yamls

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+	maxUnavailable := pdb.Spec.MaxUnavailable
+	if maxUnavailable != nil {
+		if maxUnavailable.IntVal == 0 || maxUnavailable.StrVal == "0%" {
+			result.Add(errors.ErrInvalidObject(pdb, "maxUnavailable field cannot be set to 0 or 0%"))
+		}
+	}
+
+	minAvailable := pdb.Spec.MinAvailable
+	if minAvailable != nil {
+		if minAvailable.StrVal == "100%" {
```
estroz, line 86:
```suggestion
		if minAvailable.IntVal == 100 || minAvailable.StrVal == "100%" {
```

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+// audit checks the provided rbac policies against prescribed limitations.
+// If permission is granted to create/modify a PDB, a warning is returned.
+// If permission is granted to modify default SCCs in OpenShift, an error is returned.
+func audit(policies []rbacv1.PolicyRule) (result errors.ManifestResult) {
+	// check for permission to modify/create PDBs
+	for _, rule := range policies {
+		if contains(rule.APIGroups, PodDisruptionBudgetAPIGroup) &&
+			contains(rule.Resources, "poddisruptionbudgets") &&
+			contains(rule.Verbs, rbacv1.VerbAll, "create", "update", "patch") {
+			result.Add(errors.WarnInvalidObject("RBAC includes permission to modify/create poddisruptionbudgets", rule))
```
estroz, line 158: These warnings don't say _why_ they're warnings. They should have some context, since they're just stating the obvious right now.

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	// check sccs for modifying default known SCCs
+	for _, rule := range policies {
+		if contains(rule.APIGroups, SCCAPIGroup) &&
+			contains(rule.Resources, "securitycontextconstraints") &&
+			contains(rule.Verbs, rbacv1.VerbAll, "delete", "update", "patch") &&
+			containsDefaults(rule.ResourceNames, defaultSCCs) {
+			result.Add(errors.ErrInvalidObject(rule, "RBAC includes permission to modify default SCCs"))
```
estroz, line 168: Spell out uncommon initialisms
```suggestion
			result.Add(errors.ErrInvalidObject(rule, "RBAC includes permission to modify default securitycontextconstraints"))
```

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+	"hostmount-anyuid": {},
+	"hostnetwork":      {},
+	"node-exporter":    {},
+	"nonroot":          {},
+}
+
+func validateObjects(objs ...interface{}) (results []errors.ManifestResult) {
+	for _, obj := range objs {
+		switch u := obj.(type) {
+		case *unstructured.Unstructured:
```
estroz, line 39: This is more a general comment about how validators are set up than this particular function, but the validator caller has to know what type to pass to this internal function, which isn't great. We may want to rethink the signature of the validator interface for this reason.

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,199 @@
...
+const (
+	PodDisruptionBudgetKind     = "PodDisruptionBudget"
+	PriorityClassKind           = "PriorityClass"
+	RoleKind                    = "Role"
+	ClusterRoleKind             = "ClusterRole"
+	PodDisruptionBudgetAPIGroup = "policy"
+	SCCAPIGroup                 = "security.openshift.io"
+)
+
+var defaultSCCs = map[string]struct{}{
```
estroz, line 25: Could use a comment explaining what these values pertain to.

estroz (changes requested):
Looks good overall, mostly nits.

pkg/validation/internal/object.go
exdx: If the manifest says `MinAvailable: 100` then it means they want 100 pods always available --- this is strange but not equivalent to 100% of pods available right? I don't think we need to block on the `IntVal` here, unlike in the `maxUnavailable` case.

pkg/validation/internal/object.go
exdx: That's a good point, will update

pkg/validation/internal/object.go
exdx: If `MinAvailable: 100` is on the spec it means keep at least 100 pods on the cluster -- this is strange but necessarily the same as 100% of pods. I don't think we need to block on the absolute number here, unlike in the `maxUnavailable` case where 0 and 0% are equivalent.

pkg/validation/internal/object.go
exdx: I agree. Generally all `kubectl-able` objects fulfill the `Unstructured` interface, so I think it's ok for now, but as you said it isn't flexible in what it accepts to validate.

pkg/validation/internal/object.go
exdx: I added a bit more context, but not sure if we want to add more. Open to suggestions.

pkg/validation/internal/object.go
estroz: `unstructured.Unstructured` isn't an interface, it's a type. Do you mean `runtime.Object`?

exdx: Yep, that's what I meant

I propose we keep this as-is for now and revisit the signature of the validators as part of a separate PR

pkg/validation/internal/object.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	/*
+	   maxUnavailable field cannot be set to 0 or 0%.
+	   minAvailable field cannot be set to 100%.
+	*/
+
+	maxUnavailable := pdb.Spec.MaxUnavailable
+	if maxUnavailable != nil && (maxUnavailable.IntVal == 0 || maxUnavailable.StrVal == "0%") {
+			result.Add(errors.ErrInvalidObject(pdb, "maxUnavailable field cannot be set to 0 or 0%"))
```
kevinrizza, line 81: nit: `gofmt`?

PR #63, opened 2020-09-16
Bug 1879640: (fix) csv name should not be constrained

gallettilance:


pkg/validation/internal/csv.go
```diff
@@ -50,20 +49,11 @@ func validateCSV(csv *v1alpha1.ClusterServiceVersion) errors.ManifestResult {
...
-	if len(splitName) != 2 {
-		return "", semver.Version{}, fmt.Errorf("%q must have format: {operator name}.(v)X.Y.Z", name)
-	}
-	verStr := strings.TrimLeft(splitName[1], "v")
-	nameVer, err := semver.Parse(verStr)
-	if err != nil {
-		return "", semver.Version{}, fmt.Errorf("%q contains an invalid semver %q", name, splitName[1])
-	}
-	return splitName[0], nameVer, errors.Error{}
+	return errors.Error{}
```
joelanford, line 56: Any reason we're not returning `nil` here?

joelanford (approved):
Just one nit/question about using `errors.Error{}` instead of `nil`

Otherwise,
/lgtm

pkg/validation/internal/csv.go
gallettilance: Using `nil` should be fine - was just trying to stay consistent with the rest of the code

PR #66, opened 2020-09-23
Add a "properties" field to bundle lookups.

benluddy:
If present, operator property information needs to be persisted until
bundle unpacking completes so that properties can be projected onto
ClusterServiceVersions.

This is an artifact of the fact that the on-cluster representation of an installed operator (ClusterServiceVersion) is inflated from a manifest that is contained within a bundle. The need for this may eventually be obviated by directly writing to `Operator` resources, but given that the lack of property projection is responsible for fairly serious dependency resolution problems today, I think that this medium-term solution makes sense. I have written a brief summary of property projection [here](https://hackmd.io/gSgmBLglStaHrOKJMU6BDQ?view).


ecordell (approved):
/lgtm
/approve

PR #69, opened 2020-10-19
feat: add support for custom categories

exdx:
Allow users to pass in the path to a json-encoded file of categories via the environmental variable `OPERATOR_BUNDLE_CATEGORIES` when performing optional operatorhub UI validation for CSV categories. These categories will only be used for validation if they are provided. The existing default categories will be used for validation if no custom category file is provided. 

The categories should be provided as a json encoded file with the following structure:
```
{
   "categories":[
      "Cloud Pak",
      "Registry",
      "MyCoolThing",
   ]
}
```

### DIFF
--- a/pkg/validation/internal/operatorhub.go
+++ b/pkg/validation/internal/operatorhub.go
@@ -1,9 +1,13 @@
 package internal
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/mail"
 	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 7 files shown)


pkg/validation/internal/operatorhub.go
```diff
@@ -148,12 +152,56 @@ func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) []error {
 	if categories, ok := csv.ObjectMeta.Annotations["categories"]; ok {
 		categorySlice := strings.Split(categories, ",")
 
-		for _, category := range categorySlice {
-			if _, ok := validCategories[category]; !ok {
-				errs = append(errs, fmt.Errorf("csv.Metadata.Annotations.Categories %s is not a valid category", category))
+		// use custom categories for validation if provided
+		if os.Getenv("OPERATOR_BUNDLE_CATEGORIES") != "" {
```
kevinrizza, line 156: Do we want to do the os call just once as a variable?

kevinrizza (changes requested):
Just a small comment, otherwise this looks good to me.

PR #70, opened 2020-10-27
Bug 1891898: Set min and max values on Webhook Container Port

awgreene:
Problem: Operator Authors can create CSVs that define Webhooks with
containerPorts set outside the range of 0-65535, which prevents the
ValidatingWebhookConfiguration or MutatingWebhookConfiguration from
being created due to failed validation.

Solution: Update the CSV's CRD Validation to include the proper min and
max values for webhookDescription ContainerPort field.


awgreene:
OLM logs when a CSV with an invalid containerPort is created: 

```time="2020-10-27T14:02:12Z" level=error msg="Webhooks: Error creating MutatingWebhookConfiguration: MutatingWebhookConfiguration.admissionregistration.k8s.io \"mopentelemetrycollector.kb.io-gtbpc\" is invalid: webhooks[0].clientConfig.service.port: Invalid value: 0: port is not valid: must be between 1 and 65535, inclusive"
```

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -204,7 +206,7 @@ func (w *WebhookDescription) GetValidatingWebhook(namespace string, namespaceSel
 				Name:      w.DomainName() + "-service",
 				Namespace: namespace,
 				Path:      w.WebhookPath,
-				Port:      &w.ContainerPort,
+				Port:      w.ContainerPort,
```
dinhxuanvu, line 209: Is this change intentional? It actually causes the failure on the test because of this. You will need to update the test case if that is the case.

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -228,7 +230,7 @@ func (w *WebhookDescription) GetMutatingWebhook(namespace string, namespaceSelec
 				Name:      w.DomainName() + "-service",
 				Namespace: namespace,
 				Path:      w.WebhookPath,
-				Port:      &w.ContainerPort,
```
dinhxuanvu, line 231: Ditto

dinhxuanvu (changes requested):
Looking good.

pkg/operators/v1alpha1/clusterserviceversion_types.go
awgreene: Yes - we'll want to know when this value is actually set.

pkg/operators/v1alpha1/clusterserviceversion_types.go
awgreene: Yes - we'll want to know when this value is actually set.

pkg/operators/v1alpha1/clusterserviceversion_types.go
awgreene: See above.

pkg/operators/v1alpha1/clusterserviceversion_types.go
awgreene: I suppose that we still get that information if it is not a ptr because the value will be 0.

dinhxuanvu: I meant the type change but look like you corrected on the test.

dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

PR #74, opened 2020-11-09
Introduce new OperatorCondition CRD

awgreene:
This commit introduced the OperatorCondition CRD which allows operators
to communicate advanced state directly with OLM.


pkg/operators/v1/condition_types.go
```diff
@@ -0,0 +1,39 @@
...
+type OperatorConditionStatus struct {
+	Conditions []Condition `json:"conditions,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// OperatorCondition is a Custom Resource of type `OperatorCondition` which is used to convey information to OLM about the state of an operator.
+type OperatorCondition struct {
```
awgreene, line 23: [Condition](https://github.com/operator-framework/api/blob/master/pkg/operators/v1/operator_types.go#L22) struct was already defined.

jmccormick2001 (commented):
is the user facing documentation on how to use this located in another PR?  or is it yet to be written?

anik120:
> is the user facing documentation on how to use this located in another PR? or is it yet to be written?

We should probably document this [here](https://olm.operatorframework.io/docs/concepts/crds/) and link that PR here so that we can merge both of them together/at least one right after the other.

pkg/operators/v1/condition_types.go
```diff
@@ -0,0 +1,39 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OperatorConditionSpec allows a cluster admin to convey information about the state of an operator to OLM, potentially overriding state reported by the operator.
+type OperatorConditionSpec struct {
+	Overrides []Condition `json:"overrides,omitempty"`
```
joelanford, line 9: Should we add a TODO on the `Condition` type to migrate to `metav1.Condition` when we move everything to k8s 1.19+?

Is there anything that would prevent us from doing that? (e.g. a breaking change)

awgreene: We could use the [metav1.Condition](https://github.com/kubernetes/apimachinery/blob/release-1.19/pkg/apis/meta/v1/types.go#L1367-L1413) but we should be recognize that [OLM's condition type](https://github.com/operator-framework/api/blob/master/pkg/operators/v1/operator_types.go#L21-L39) has different fields:
* OLM's Condition has the `LastUpdateTime` field.
* The metav1.Condition has the `ObservedGeneration` field.

If we plan to update this struct to use the `metav1.Condition` we will have to handle any CRs that have set the `LastUpdateTime` field. Thoughts?

njhale: I would definitely opt for using `metav1.Condition` directly. We should diverge as little as possible. If we find that `LastUpdateTime` is actually useful, then we can create a composite struct that embeds `metav1.Condition`.

pkg/operators/v1/condition_types.go
```diff
@@ -0,0 +1,39 @@
...
+	Status OperatorConditionStatus `json:"status"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// ConditionList represents a list of OperatorConditions.
+type OperatorConditionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	Items []Condition `json:"items"`
```
njhale, line 38:
```suggestion
	Items []OperatorCondition `json:"items"`
```

pkg/operators/v1/condition_types.go
njhale: to add to this, I was planning on replacing the `Condition` struct you're using w/ `metav1.Condition` at some point down the road as well.

crds/operators.coreos.com_operatorconditions.yaml
```diff
@@ -0,0 +1,91 @@
...
+spec:
+  group: operators.coreos.com
+  names:
+    kind: OperatorCondition
+    listKind: OperatorConditionList
+    plural: operatorconditions
+    singular: operatorcondition
+  scope: Namespaced
+  versions:
+  - name: v1
```
exdx, line 17: Is this really a v1 API or should we keep it in line with the other OLM types as v1alpha1? v1 is effectively saying it will never change

awgreene: Until we have an upstream/downstream split new CRDs need to be introduced at v1, this is an openshift requirements born from the fact that we offer higher levels of support than guaranteed by lower api versions.

dinhxuanvu (commented):
/lgtm

crds/operators.coreos.com_operatorconditions.yaml
exdx: Interesting -- Bit paradoxical that our oldest and most stable apis are at alpha and the newest apis are at v1 but I see the reasoning.

crds/operators.coreos.com_operatorconditions.yaml
```diff
@@ -0,0 +1,190 @@
...
+  creationTimestamp: null
+  name: operatorconditions.operators.coreos.com
+spec:
+  group: operators.coreos.com
+  names:
+    kind: OperatorCondition
+    listKind: OperatorConditionList
+    plural: operatorconditions
+    singular: operatorcondition
+  scope: Namespaced
```
dinhxuanvu, line 15: I wonder if we consider this CRD to be cluster scoped? It will align with Operator API that way as well. Maybe that is not the intention here. Just curious.

dinhxuanvu (approved):
/lgtm

crds/operators.coreos.com_operatorconditions.yaml
awgreene: Thanks for bringing this up @dinhxuanvu, I had thought about this as well and I would love a sanity check on my reasoning.

Currently OLM allows multiple installs of the same operator provided the scopes do not intercept. I don't believe that one scoped operator should prevent a different scoped operator to upgrade. By having the CR namespaced, we can continue to support scoped upgrades.

If OLM ever deprecates OperatorGroups and all operator are global, then the namespaced operatorCondition could still be placed in the namespace that the operator pod is deployed to.

awgreene: I suppose we could accomplish the same behavior by using lables and associating the OperatorCondition with a specific operator, I would love to hear if @kevinrizza and @ecordell have a preference.

awgreene: >I suppose if we wanted the resource to be cluster scoped we could accomplish the same behavior by using lables and associating the OperatorCondition with a specific operator, I would love to hear if @kevinrizza and @ecordell have a preference.

The only negative experience here is that scoped operators now require the ability to write to a non-namespaced resource, which might ruffle some feathers.

dinhxuanvu: I can see your reasoning there. At the moment, operator is pretty much namespaced so it makes sense to have namspaced conditions. Operator API can aggregate Condition status just like any other OLM objects so it is fine to go this way. However, I thought there is a synergy between OperatorCondition and Operator API. I think cluster-scoped OperatorCondition can be used as a signal for an operator to know if the other operator that it may depend on is available or not. It can be useful that way. I'm not 100% sure though. Anyway, I certainly don't want to cause a scope creep with this comment.

joelanford (approved):
/lgtm

PR #77, opened 2020-11-20
Update github action setup-go to version 1.1.3     

awgreene:
Fixes a security vulnerability that allowed environment variables and path injection in workflows that log untrusted data to STDOUT, possibly resulting in environment variables being introducd or modified without the intention of the workflow operator.
    
Ref: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


dinhxuanvu (approved):
/lgtm

PR #79, opened 2020-11-24
Update CatalogSource Spec to include ImageSecret

anik120:
Currently there is no clean way to attach a secret to a CatalogSource
that can be used to pull down index images from private registries.
Ref [this issue](https://github.com/operator-framework/operator-lifecycle-manager/issues/1505).
This PR introduces spec.imageSecret to v1alpha1.CataglogSource so that
a secretName can be specified to be used while pulling down the index
image.


crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -86,10 +86,18 @@ spec:
                   mediatype:
                     type: string
               image:
-                description: Image is an operator-registry container image to instantiate
-                  a registry-server with. Only used when SourceType = SourceTypeGrpc.
+                description: Image is a container image of an index which hosts a
+                  collection of operator bundles. Only used when SourceType = SourceTypeGrpc.
                   If present, the address field is ignored.
                 type: string
+              imageSecret:
```
njhale, line 93: I don't think this disambiguates secret usage as much as treating access to a catalog and its contents as a whole would; i.e. giving access to a catalog requires giving access to pull its image and all images that it references. In this world, we have one set of secrets referenced by the CatalogSource object, which are automatically bound to the default serviceaccount of the catalog pods as well as copied and automatically bound to operator serviceaccounts. Having two fields makes for a more complex UX. I also don't really see the case for segregating access between an index and its content, at that point why not just have separate catalogs for private content. We also played around with this concept with `CatalogSourceConfigs`, which we have since abandoned.

Further, I posit that it may have been an oversight that secrets weren't bound to the catalogsource pod's serviceaccount  to begin with.

njhale (changes requested):
After reading through the source issues in detail, I still hold my reservations about adding a new field here. I am also being cautious because each new field added is something that we need to support for quite some time.

crds/operators.coreos.com_catalogsources.yaml
anik120: >I don't think this disambiguates secret usage as much as treating access to a catalog and its contents as a whole would; i.e. giving access to a catalog requires giving access to pull its image and all images that it references.

>I also don't really see the case for segregating access between an index and its content, at that point why not just have separate catalogs for private content.

These are assuming that the index author has permission to give access to every bundle image s/he is including in the index. 
Consider this scenario: I am a private index author, who wants to include an operator with a bundle image that's public, and another operator who's bundle image requires a pull secret. 
Bottom line: Separation of concern - index image is one thing, and bundle images are another, and it's probably presumptuous to require users to give access to everything with just one pull secret to rule them all.  

 > In this world, we have one set of secrets referenced by the CatalogSource object, which are automatically bound to the default serviceaccount of the catalog pods as well as copied and automatically bound to operator serviceaccounts.

Quoting @exdx as my response: 

"@nhale’s suggestion to bind the secret to the default SA in the catalog pod’s namespace also makes sense, although it makes the assumption that OLM should always use the default SA  to pull down catalog images. I believe the default SA is always present in a namespace but one can imagine it can be heavily restricted/disabled. We don’t provide a way for a user to specify which SA should be used in the catalog pod."
 
> Having two fields makes for a more complex UX

I am of the opinion that `spec.secrets` might have made sense in the previous world,  but in this world where secrets could be for a variety of categories of images, i.e index image or bundle image etc, `spec.secrets` just doesn't convey anything and should be renamed too. But we could start with 
```
spec:
       image: bar/foo/image
       imageSecret: 
                 - secretName: baz
```

> I am also being cautious because each new field added is something that we need to support for quite some time.

Could you expand on how the support for this might get complex?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -38,6 +38,10 @@ const (
 	CatalogSourceRegistryServerError ConditionReason = "RegistryServerError"
 )
 
+type IndexImageSecret struct {
+	// SecretName is the name of the kubernetes Secret object.
+	SecretName string `json:"secretName,omitempty"`
```
exdx, line 43: IMO this field would be a bit repetitive on the spec, with something like 
```
secretName:
  imageSecret: <my-secret-name>
```
would something more generic like `auth` on the spec be more extensible? 
```
auth:
   catalogPullSecret: <my-secret-name>
```

crds/operators.coreos.com_catalogsources.yaml
exdx: > I don't think this disambiguates secret usage as much as treating access to a catalog and its contents as a whole would; i.e. giving access to a catalog requires giving access to pull its image and all images that it references.
> 

This is very true, we should probably treat the access to the catalog itself and its content as the same. In that case the SA approach makes more sense.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: @exdx I was trying to come up with something better than those two too. 

FYI right now it's not 

```
secretName: 
    imageSecret: <my-secret-name>
```

It's 
```
imageSecret: 
     secretName: <my-secret-name>
```
But something like 
```
auth:
     imageSecret: <my-secret-name>
```
maybe more intuitive? 
`catalogPullSecret` may sound like it's asking the user to plug in the registry account pull secret there instead of mentioning the secret name.

dmesser:
In general it seems we have two concerns:

1) authenticating against a registry which holds the catalog index image

2) authenticating against a registry which holds the operator and operand and related images

Because the owner of the catalog could be different from the owner the operator, operand and related images we should be able to cater for this use case (Operator catalog curation mainly). if we don't, wouldn't we leak catalog index image secrets into all namespaces where the Operator is deployed?

PR #81, opened 2020-12-01
Add OperatorUpgradeable condition type to OperatorCondition api

dinhxuanvu:
This condition type is used to signify that the operator is
upgradeable.


ecordell (approved):
/approve

PR #85, opened 2020-12-03
Change OperatorUpgradeable type to string and add utils func

dinhxuanvu:
Use string as OperatorUpgradeable type instead of intermediate
type ConditionType.

Add utils func FindConditionType to look up Condition type.


awgreene (approved):
/lgtm

PR #87, opened 2020-12-08
add shortName for operatorcondition

jianzhangbjz:
There is no `condition` described in https://github.com/operator-framework/olm-docs/pull/83, so add the `condition` as the short name of the `operatorcondition`
```
[root@preserve-olm-env data]#  oc get condition
error: the server doesn't have a resource type "condition"

[root@preserve-olm-env data]#  oc get operatorcondition
No resources found in default namespace.
```


crds/operators.coreos.com_operatorconditions.yaml
```diff
@@ -11,6 +11,8 @@ spec:
     kind: OperatorCondition
     listKind: OperatorConditionList
     plural: operatorconditions
+    shortNames:
+    - condition
```
awgreene, line 15: This will be removed when the next version of the CRD is generated. Can you append a shortname annotation to the [go struct](https://github.com/operator-framework/api/blob/15d24f1a8352499de54d5efff97dd8b885fc1359/pkg/operators/v1/operatorcondition_types.go#L30)?

Example: https://github.com/operator-framework/api/blob/15d24f1a8352499de54d5efff97dd8b885fc1359/pkg/operators/v1/operatorgroup_types.go#L61

awgreene: Also, make sure to run `make manifests` in the root of the repository after making this change.

awgreene (changes requested):
Nice work @jianzhangbjz - one request to make sure that the change isn't removed when CRDs are generated.

crds/operators.coreos.com_operatorconditions.yaml
jianzhangbjz: @awgreene Sorry for the late reply, got it, thanks very much! Updated it.

jianzhangbjz:
@benluddy Could you help have a review? Thanks!

jianzhangbjz:
Both `approved` and `lgtm` labels added, but this PR still not be merged, anything else needed?

PR #95, opened 2021-01-28
fix error on missing icon

gallettilance:


jmrodri:
Using this patch in the SDK seems to fix the error:
```
$ /home/devel/jesusr/dev/operator-sdk/build/operator-sdk  bundle validate openshift-ose-clusterresourceoverride-operator-bundle:v4.6.0
INFO[0000] Unpacking image layers                       
INFO[0000] running /usr/bin/docker pull registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-clusterresourceoverride-operator-bundle:v4.6.0.202101230113.p0-1 
INFO[0000] running docker create                        
INFO[0000] running docker cp                            
INFO[0000] running docker rm                            
INFO[0000] Found annotations file                        bundle-dir=/tmp/bundle-977913200 container-tool=docker
INFO[0000] Could not find optional dependencies file     bundle-dir=/tmp/bundle-977913200 container-tool=docker
INFO[0000] All validation tests have completed successfully 
```

pkg/validation/internal/operatorhub.go
```diff
@@ -74,16 +74,19 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	for _, warn := range warns {
+		result.Add(errors.WarnInvalidCSV(warn.Error(), bundle.CSV.GetName()))
+	}
 	return result
 }
 
-func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) []error {
+func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) ([]error, []error) {
 	var errs []error
+	var warns []error
```
kevinrizza, line 89: This seems like the wrong type to use? We're not overloading any of the error methods here at all (or implementing anything special on top), and we're ultimately just calling warn.Error() to get the string back out. Seems like this could just be a slice of strings?

jmrodri: @kevinrizza it looks like he was just reusing what was already there. The logic assumes a slice of errors. But what you suggest is more correct because the strings are passed into specific `Error`s. +1 to using `[]string` but will that mess anything else up?

PR #97, opened 2021-02-22
Make subscription spec.config and spec.config.resources optional

joelanford:
Required by operator-framework/operator-lifecycle-manager#2010

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>


joelanford:
This is a breaking change for Go users of `operator-framework/api`.

As far as I can tell, this is NOT a breaking change from a Kubernetes API standpoint. The current CRD does not require [`spec.config`](https://github.com/operator-framework/api/blob/bf75ca75fea1d28b4b38f88487eca3b8bbefd6c1/crds/operators.coreos.com_subscriptions.yaml#L55-L58) or `spec.config.resources` (which has no required fields).

joelanford:
There's likely a way to fix this on the OLM side without making an API change by treating an empty struct identically to a nil struct pointer. However that has the downside that it becomes impossible to use `spec.config.resources` to unset resources defined by the CSV deployment.

pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -29,13 +29,13 @@ const (
...
-	StartingCSV            string             `json:"startingCSV,omitempty"`
-	InstallPlanApproval    Approval           `json:"installPlanApproval,omitempty"`
-	Config                 SubscriptionConfig `json:"config,omitempty"`
+	CatalogSource          string              `json:"source"`
+	CatalogSourceNamespace string              `json:"sourceNamespace"`
+	Package                string              `json:"name"`
+	Channel                string              `json:"channel,omitempty"`
+	StartingCSV            string              `json:"startingCSV,omitempty"`
+	InstallPlanApproval    Approval            `json:"installPlanApproval,omitempty"`
+	Config                 *SubscriptionConfig `json:"config,omitempty"`
```
joelanford, line 38: This change is not directly related to bug fix with overriding resources. However it seems that the intent is for `spec.config` to be optional, so it seems like a reasonable time to make this change as well.

dinhxuanvu (approved):
/lgtm

PR #98, opened 2021-02-23
Add optional go-apidiff test to CI

joelanford:
I wrote [`go-apidiff`](https://github.com/joelanford/go-apidiff) to help maintainers of `kubernetes-sigs/controller-runtime` easily identify PRs that break Go APIs. It's also being used in `kubernetes-sigs/cluster-api*` repos.

Today I finally got around to making a GitHub action out of it. I figured it might help in some of the operator-framework repos that are used as libraries.

It will compare the PR head with the target branch, report any API changes, and fail if any of them are breaking changes. Other projects using this make this test optional so that it is a simple signal to maintainers about whether changes are API-compatible.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -11,12 +11,29 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-    - name: Set up Go 1.15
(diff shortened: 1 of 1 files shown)


hasbro17 (approved):
/lgtm

PR #99, opened 2021-02-25
Expand validation to annotations

jchunkins:
This is the first step in implementing issue https://github.com/operator-framework/operator-registry/issues/568

I started with the annotations that I knew were problematic (e.g. using wrong case `olm.skiprange` when the annotation should be `olm.skipRange`) and also searched for other case sensitive annotations I could find. I also expanded the validation types to include OperatorGroups as well since it can also use case sensitive annotation names. Note that OperatorGroups are not directly hooked up to the command line execution of `operator-verify` (similar to how `PackageManifestValidator` is not accessible from the command line either).

I intend to open up new PRs in operator-registry once this is merged here so we can pick up the changes for CSV validation.

- Add validation function for well known case sensitive annotation names
- Use new validation function in CSV and OperatorGroups
- Add minimal OperatorGroup validator calling new function
- Add / update unit tests and test data
- update usage readme

Signed-off-by: John Hunkins <jhunkins@us.ibm.com>

### DIFF
--- a/pkg/validation/internal/annotations.go
+++ b/pkg/validation/internal/annotations.go
@@ -1,0 +1,53 @@
+package internal
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/validation/errors"
+)
+
+// CaseSensitiveAnnotationKeySet is a set of annotation keys that are case sensitive
+// and can be used for validation purposes. The key is always lowercase and the value
+// contains the expected case sensitive string. This may not be an exhaustive list.
+var CaseSensitiveAnnotationKeySet = map[string]string{
+
+	strings.ToLower(v1.OperatorGroupAnnotationKey):             v1.OperatorGroupAnnotationKey,
+	strings.ToLower(v1.OperatorGroupNamespaceAnnotationKey):    v1.OperatorGroupNamespaceAnnotationKey,
+	strings.ToLower(v1.OperatorGroupTargetsAnnotationKey):      v1.OperatorGroupTargetsAnnotationKey,
+	strings.ToLower(v1.OperatorGroupProvidedAPIsAnnotationKey): v1.OperatorGroupProvidedAPIsAnnotationKey,
+	strings.ToLower(v1alpha1.SkipRangeAnnotationKey):           v1alpha1.SkipRangeAnnotationKey,
+}
+
+/*
+ValidateAnnotationNames will check annotation keys to ensure they are using
+proper case. Uses CaseSensitiveAnnotationKeySet as a source for keys
+which are known to be case sensitive. This function can be used anywhere
+annotations need to be checked for case sensitivity.
+
+Arguments
+
+• annotations: annotations map usually obtained from ObjectMeta.GetAnnotations()
+
+• value: is the field or file that caused an error or warning
+
+Returns
+
+• errs: Any errors that may have been detected with the annotation keys provided
+*/
+func ValidateAnnotationNames(annotations map[string]string, value interface{}) (errs []errors.Error) {
+	// for every annotation provided
+	for annotationKey := range annotations {
+		// check the case sensitive key set for a matching lowercase annotation
+		if knownCaseSensitiveKey, ok := CaseSensitiveAnnotationKeySet[strings.ToLower(annotationKey)]; ok {
+			// we have a case-insensitive match... now check to see if the case is really correct
+			if annotationKey != knownCaseSensitiveKey {
+				// annotation key supplied is invalid due to bad case.
+				errs = append(errs, errors.ErrFailedValidation(fmt.Sprintf("provided annotation %s uses wrong case and should be %s instead", annotationKey, knownCaseSensitiveKey), value))
+			}
+		}
+	}
+	return errs
+}
--- a/README.md
+++ b/README.md
@@ -17,4 +17,4 @@ You can install the `operator-verify` tool from source using:
 
 To verify your ClusterServiceVersion yaml,
 
-`$ operator-verify verify /path/to/filename.yaml`
+`$ operator-verify manifests /path/to/filename.yaml`
--- a/pkg/operators/v1alpha1/clusterserviceversion_types.go
+++ b/pkg/operators/v1alpha1/clusterserviceversion_types.go
@@ -22,6 +22,7 @@ const (
 	ClusterServiceVersionKind           = "ClusterServiceVersion"
 	OperatorGroupNamespaceAnnotationKey = "olm.operatorNamespace"
 	InstallStrategyNameDeployment       = "deployment"
+	SkipRangeAnnotationKey              = "olm.skipRange"
 )
 
 // InstallModeType is a supported type of install mode for CSV installation
--- a/pkg/validation/internal/csv.go
+++ b/pkg/validation/internal/csv.go
@@ -46,6 +46,8 @@ func validateCSV(csv *v1alpha1.ClusterServiceVersion) errors.ManifestResult {
 	result.Add(validateInstallModes(csv)...)
 	// check missing optional/mandatory fields.
 	result.Add(checkFields(*csv)...)
+	// validate case sensitive annotation names
+	result.Add(ValidateAnnotationNames(csv.GetAnnotations(), csv.GetName())...)
 	return result
 }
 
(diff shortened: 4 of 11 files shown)


pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,65 @@
...
+
+Returns
+
+Any errors that may have been detected with the annotation keys provided
+*/
+func ValidateAnnotationNames(annotations map[string]string, value interface{}) (errs []errors.Error) {
+	// for every annotation provided
+	for annotationKey := range annotations {
+		// loop through all well known keys
+		for knownCaseSensitiveKey := range CaseSensitiveAnnotationKeySet {
```
dinhxuanvu, line 50: You create `CaseSensitiveAnnotationKeySet` as a map. Why don't you use the key look up to check if the key exists or not? It is faster. No need to loop through the entire map like this.

pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,65 @@
...
+			// case insensitively compare this annotation key with this specific well known key.
+			// we want to determine if the key was intended to be used
+			if strings.EqualFold(annotationKey, knownCaseSensitiveKey) {
+				// we have a case-insensitive match... now check to see if the case is really correct
+				if annotationKey != knownCaseSensitiveKey {
+					// annotation key supplied is invalid due to bad case.
+					errs = append(errs, errors.ErrFailedValidation(fmt.Sprintf("provided annotation %s uses wrong case and should be %s instead", annotationKey, knownCaseSensitiveKey), value))
+				}
+				// bail out... we found and checked what we're looking for with this annotationKey
+				break
```
dinhxuanvu, line 60: Should we check every annotations and project the errors for all of them instead of bailing out as soon as we find one problem?

dinhxuanvu (commented):
Lookin good. Just a few comments.

pkg/validation/internal/annotations.go
dinhxuanvu: On the second look, I can see why you don't use hash look-up because of the case-sensitive nature. See my comment on the map variable.

pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,65 @@
...
+
+// CaseSensitiveAnnotationKeySet is a set of annotation keys that are case sensitive
+// and can be used for validation purposes. This may not be an exhaustive list.
+var CaseSensitiveAnnotationKeySet = map[string]struct{}{
+
+	v1.OperatorGroupAnnotationKey:             {},
+	v1.OperatorGroupNamespaceAnnotationKey:    {},
+	v1.OperatorGroupTargetsAnnotationKey:      {},
+	v1.OperatorGroupProvidedAPIsAnnotationKey: {},
+	"olm.skipRange":                           {},
```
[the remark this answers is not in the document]
jchunkins: Yes. I can do that. I'll add a constant to `pkg/operators/v1alpha1/clusterserviceversion_types.go`

pkg/validation/internal/annotations.go
jchunkins: I will be reimplementing this given your other comment, but I'll add a response just so I've addressed this comment:

The `break` in this case would only affect the inner loop where we are ranging on `CaseSensitiveAnnotationKeySet` and only when we've established that `annotationKey` matches case insensitively with the `knownCaseSensitiveKey`. In other words, the current `annotationKey` **is** a known case sensitive key (but we don't yet know if it matches in a case sensitive manner yet), so there's no point looking at any other known keys. The remaining annotations provided would continue to be checked when the outer loop continues on.

pkg/validation/internal/annotations.go
jchunkins: I plan to reimplement based on your other comment.

pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"strings"
+
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/api/pkg/validation/errors"
+)
+
+// CaseSensitiveAnnotationKeySet is a set of annotation keys that are case sensitive
+// and can be used for validation purposes. This may not be an exhaustive list.
+var CaseSensitiveAnnotationKeySet = map[string]struct{}{
```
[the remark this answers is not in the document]
jchunkins: I've reimplemented based on this comment. So the lookup will be faster, but you still need to use strings.lowercase() on the current annotation key before you can use it in the map lookup. So you gain performance improvement in the lookup, but have the cost of the strings.lowercase() now. Given the numbers of annotations to check this won't be noticeable one way or the other.

Bottom line your way makes the code cleaner. :)

pkg/validation/internal/annotations.go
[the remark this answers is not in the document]
jchunkins: Fixed in new commit

pkg/validation/internal/annotations.go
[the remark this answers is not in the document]
jchunkins: fixed in new commit

dinhxuanvu (approved):
/lgtm

pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,61 @@
...
+	strings.ToLower(v1.OperatorGroupProvidedAPIsAnnotationKey): v1.OperatorGroupProvidedAPIsAnnotationKey,
+	strings.ToLower(v1alpha1.SkipRangeAnnotationKey):           v1alpha1.SkipRangeAnnotationKey,
+
+	// below are duplicate values of known keys
+
+	// strings.ToLower(v1alpha1.OperatorGroupNamespaceAnnotationKey):    v1alpha1.OperatorGroupNamespaceAnnotationKey,
+	// strings.ToLower(v1alpha2.OperatorGroupAnnotationKey):             v1alpha2.OperatorGroupAnnotationKey,
+	// strings.ToLower(v1alpha2.OperatorGroupNamespaceAnnotationKey):    v1alpha2.OperatorGroupNamespaceAnnotationKey,
+	// strings.ToLower(v1alpha2.OperatorGroupTargetsAnnotationKey):      v1alpha2.OperatorGroupTargetsAnnotationKey,
+	// strings.ToLower(v1alpha2.OperatorGroupProvidedAPIsAnnotationKey): v1alpha2.OperatorGroupProvidedAPIsAnnotationKey,
```
awgreene, line 29: It's probably safe to remove these.

pkg/validation/internal/annotations.go
```diff
@@ -0,0 +1,61 @@
...
+
+Arguments
+
+• annotations: annotations map usually obtained from ObjectMeta.GetAnnotations()
+
+• value: is the field or file that caused an error or warning
+
+Returns
+
+Any errors that may have been detected with the annotation keys provided
```
awgreene, line 46: Non-blocking nit: I am unfamiliar with the format, but based on the Arguments above should it match:
```suggestion
• errs: Any errors that may have been detected with the annotation keys provided
```

awgreene (approved):
Great work @jchunkins, left some non-blocking nits.

pkg/validation/internal/annotations.go
jchunkins: I have removed these.

pkg/validation/internal/annotations.go
jchunkins: accepted suggested change

awgreene (approved):
/lgtm

njhale (commented):
Ah, this looks interesting! Although it does seem fairly limited; i.e. in order to catch similar mistakes (e.g. `olm.OperatorGroups`, `olm.operator-group`, etc), we need to add more custom logic.

Have you considered something more generalized? maybe this could use a string distance algorithm (see [agnivade/levenschtein][1]) to catch non-exact matches and suggest the closest correct key?

[1]: https://github.com/agnivade/levenshtein

jchunkins:
@njhale I appreciate the idea and can see how that can make this more useful. However, considering a crawl/walk/run approach, maybe we start off with this implementation before committing to additional complexity. 

My primary concern would be introducing false positives when you start handling more complex scenarios. A secondary concern would be the computation time cost involved during validation (which admittedly might not be that large). Right now we know that we have real world examples of folks making simple mistakes with case sensitivity being the root problem.

benluddy (commented):
Thanks for your contribution! I don't have a problem with a specific case-sensitivity rule like this one.

It might eventually be useful to validate that any annotation key beginning with "olm." (?) can be found on an annotation allowlist, which would cover the other mistakes Nick brought up. That would take a bit of extra effort to offer a "did you mean" like the one your patch provides, though.

PR #100, opened 2021-03-04
Add API for cleanup spec and status

hasbro17:
This PR introduces the operand cleanup API that allows configuring the CSV to enable cleanup of CRs managed by this CSV's operator when the CSV is deleted.

This also includes a status.cleanup block that will be used to show the progress of cleanup.

Enhancement: https://github.com/operator-framework/enhancements/blob/master/enhancements/olm-deletes-operands.md
RFE: https://github.com/operator-framework/operator-lifecycle-manager/issues/1787

**TODO:**
- We'll update the cleanup spec/status APIs as required by the implementation and then merge once all requirements are fulfilled.
- See: https://github.com/operator-framework/operator-lifecycle-manager/pull/2030


pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -495,6 +503,59 @@ type ClusterServiceVersionStatus struct {
...
+	// PendingDeletion is the list of custom resource objects that are pending deletion and blocked on finalizers.
+	// This indicates the progress of cleanup that is blocking CSV deletion or operator uninstall.
+	// +optional
+	PendingDeletion []ResourceList `json:"pendingDeletion,omitempty"`
+}
+
+// ResourceList represents a list of resources which are of the same GVK
+type ResourceList struct {
+	Group     string           `json:"group"`
+	Version   string           `json:"version"`
```
hasbro17, line 522: Note that this is a slight departure [from the proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/olm-deletes-operands.md#operand-cleanup-status) which was not keying the CR list by GVK. This makes the list slightly less verbose.

Example of what the status block currently looks like
```YAML
status:
  cleanup:
    pendingDeletion:
      - group: etcd.database.coreos.com
        instances:
          - name: cluster-1
            namespace: default
          - name: cluster-2
            namespace: openshift-operators
        kind: EtcdCluster
        version: v1beta2
      - group: etcd.database.coreos.com
        instances: []
        kind: EtcdBackup
        version: v1beta2
...
```

However we probably only need to key the CR list by GK instead of GVK.

If a CSV supports an ownedCRD of the same GK but two different versions (with conversion webhooks implemented if needed) then the underlying stored objects are the same even if we can query them at different versions.
```YAML
spec:
  customresourcedefinitions:
    owned:
      - description: Represents a cluster of etcd nodes.
        displayName: etcd Cluster
        kind: EtcdCluster
        name: etcdclusters.etcd.database.coreos.com
        version: v1beta2
        ...
      - description: Represents a cluster of etcd nodes.
        displayName: etcd Cluster
        kind: EtcdCluster
        name: etcdclusters.etcd.database.coreos.com
        version: v1
        ...
```


So we probably don't want to display the same CR at different versions in the cleanup status e.g:

```YAML
status:
  cleanup:
    pendingDeletion:
      - group: etcd.database.coreos.com
        kind: EtcdCluster
        version: v1beta2
        instances:
          - name: cluster-1
            namespace: default
      - group: etcd.database.coreos.com
        kind: EtcdCluster
        version: v1
        instances:
          - name: cluster-1
            namespace: default
...
```

hasbro17:
/hold

Should be okay to review but we might have to make adjustments depending on the implementation at https://github.com/operator-framework/operator-lifecycle-manager/pull/2030

pkg/operators/v1alpha1/clusterserviceversion_types.go
exdx: Removed the version field from the list

PR #102, opened 2021-03-17
Add cuelang schema for JSON config validation

dinhxuanvu:
cuelang schema will be used to validate JSON config


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/operator-framework/api
 go 1.15
 
 require (
+	cuelang.org/go v0.2.2
```
ecordell, line 6: 0.3 is much more stable, I would pull the latest release here

pkg/config/schema/schema.cue
```diff
@@ -0,0 +1,99 @@
...
+  icon: #icon
+  description: !=""
+  properties?: [...#property]
+  ...
+}
+
+// schema: "olm.bundle"
+#olmbundle: {
+  schema: "olm.bundle"
+  name: !=""
```
ecordell, line 87: a lot of these `!=""` should probably specify types as well, this should be a string too

pkg/config/schema/schema.cue
```diff
@@ -0,0 +1,99 @@
...
+  defaultChannel: !=""
+  icon: #icon
+  description: !=""
+  properties?: [...#property]
+  ...
+}
+
+// schema: "olm.bundle"
+#olmbundle: {
```
ecordell, line 85: I think we want to make use of the metaschema here, i.e.

```
#olmbundle: #blob & {}
```

pkg/config/schema/schema.cue
```diff
@@ -0,0 +1,99 @@
...
+  name: !=""
+  image: !=""
+  properties: [...#property]
+  relatedImages: [...#relatedImages]
+  ...
+}
+
+#item: {
+  schema: !=""
+  name: !=""
```
ecordell, line 96: I don't think `name` should be a common field to all blobs

pkg/config/schema/schema.cue
```diff
@@ -0,0 +1,99 @@
...
+  type: "olm.channel"
+  value: #channel
+}
+
+#olmskips: {
+  type: "olm.skips"
+  value: !=""
+}
+
+#olmskipRange: {
```
ecordell, line 57: all of the properties should be `#property & { more specific schema }`

ecordell (commented):
looking good! left a few comments mostly about the cue

go.mod
dinhxuanvu: Already changed this locally. Haven't pushed it up.

pkg/config/schema/schema.cue
dinhxuanvu: I think `!=""` only works with string. Plus, I'm still having problem with json validation library in cuelang doesn't honor this non-empty check.

dinhxuanvu: I opened a PR on cue repo to address this bug: https://github.com/cuelang/cue/pull/853

pkg/config/schema/schema.cue
```diff
@@ -0,0 +1,96 @@
+package config
```
njhale, line 1: we keep calling this "config" but we're really defining the schema of a catalog/index, right?

pkg/config/validation/testdata/invalid/bundle.json
```diff
@@ -0,0 +1,32 @@
...
+                "name": "alpha"
+            }
+        }
+    ],
+    "relatedImages": [
+        {
+            "image": "quay.io/coreos/etcd-operator@sha256:bd944a211eaf8f31da5e6d69e8541e7cada8f16a9f7a5a570b22478997819943"
+        }
+    ]
+}
```
njhale, line 32: consider using something like https://golang.org/pkg/testing/quick/ to write property based/generative tests as well

pkg/config/validation/validate.go
```diff
@@ -0,0 +1,51 @@
...
+	olmPackageRequired = "olmpackagerequired"
+)
+
+type configValidator struct {
+	runtime  *cue.Runtime
+	instance *build.Instance
+	logger   *logrus.Entry
+}
+
+func (c configValidator) Validate(b []byte, key string) error {
```
njhale, line 31: Can we expect clients to:
- always know the key of the blob they want to validate?
- call this method with every key to check all schema types?

maybe I'm missing something; does `key: ""` check everything?

pkg/config/validation/validate_test.go
```diff
@@ -0,0 +1,122 @@
...
+			hasError:    true,
+			errString:   `#olmbundle.relatedImages.0.name: incomplete value !=""`,
+		},
+		{
+			description: "invalid channel property config",
+			filename:    "./testdata/invalid/channel_property.json",
+			kind:        olmChannel,
+			hasError:    true,
+			errString:   `#olmchannel.value.name: conflicting values !="" and 1 (mismatched types string and int)`,
+		},
```
njhale, line 48: this makes me wonder if you can the validator will pass any valid top-level channel_property

njhale (commented):
Looking really good. Nicely organized!

A few comments/questions.

pkg/config/schema/schema.cue
dinhxuanvu: Correct. Perhaps, we can call this pkg `schema`.

pkg/config/validation/validate.go
dinhxuanvu: The key is the name of the schema definition (`#something` that you see on the cue file) that you want to validate the JSON blob against. My intention is to let registry to decide the JSON format and let it be responsible for parsing the JSON and "dissect" into pieces.  Then, it will feed pieces into this library to validate them against a specific definition in the default schema or custom schema. I want this part to be configurable based on inputs so carrying the knowledge of config JSON structure into this is not my desire.

njhale: ah, I see.

I was thinking it would be nice to be able to execute `cue vet my-index-config.json schema.cue` to validate an entire index in one go.

dinhxuanvu: Correct, registry needs to have its own so-called schema to parse config. Registry dictates the schema actually. The schema you see under `/manifests` is the default one. The validation command should accept a custom schema if requested and this library will validate against the definitions specified in custom schema instead of the default one.

Even with the `cue vet` command, you will still need to pick a definition in the cue file to validate again the json file. It won't automatically go through the list of definitions and run validation against each of them.

You can still validate top-level `olm.bundle` and `olm.package` blob and the validation will perform the check in all of specs that are nested underneath the top-level such as `properties` and `relatedImages`. However, because the property is not homogeneous AKA we have variety of properties that are different (different fields or different type for the same field) depending on which type it is, the validation is unable to validate individual property type at the top-level. It can only validate the generic schema AKA the fields that are shared between all of properties.

PR #103, opened 2021-03-23
Add new checks for operatorhub validator

camilamacedo86:
**Description**
- Implementation of the EP: https://github.com/operator-framework/enhancements/blob/master/enhancements/operator-hub-check.md
- All new tests/checks are covered with tests
- Encapsulate the checks and add comments to make it easier to keep them maintained. 
PS.: Also, we increase as well as the coverage of the pre-existent ones such as the icon check

**See the changes done here tested/checked with SDK**

Screen Shot 2021-03-24 at 01 43 43
Screen Shot 2021-03-24 at 01 44 06

See the PR: https://github.com/operator-framework/operator-sdk/pull/4693 (using the fork)


pkg/validation/internal/operatorhub.go
```diff
@@ -20,48 +20,58 @@ import (
...
-	"Developer Tools":        struct{}{},
-	"Database":               struct{}{},
-	"Integration & Delivery": struct{}{},
-	"Logging & Tracing":      struct{}{},
-	"Monitoring":             struct{}{},
-	"Networking":             struct{}{},
-	"OpenShift Optional":     struct{}{},
-	"Security":               struct{}{},
-	"Storage":                struct{}{},
-	"Streaming & Messaging":  struct{}{},
```
camilamacedo86, line 51: just a cleanup since it would no pass in a lint check.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +91,186 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	var warns []error
+
+	errs = checkSpecProviderName(csv, errs)
+	errs = checkSpecMaintainers(csv, errs)
+	errs = checkSpecLinks(csv, errs)
+	errs, warns = checkAnnotations(csv, errs, warns)
+	errs = checkSpecVersion(csv, errs)
+	errs, warns = checkSpecIcon(csv, errs, warns)
+	errs, warns = checkSpecMinKubeVersion(csv, warns, errs)
+	errs, warns = checkBundleName(csv, warns, errs)
```
camilamacedo86, line 216: See that all check still the same they are only encapsulated to reduce the complexity, be easier we cover then with the tests as to keep them maintained.

pkg/validation/internal/operatorhub_test.go
```diff
@@ -129,46 +130,331 @@ func TestExtractCategories(t *testing.T) {
...
+		name        string
+		args        args
+		wantError   bool
+		wantWarning bool
 		errStrings  []string
 		warnStrings []string
 	}{
 		{
-			description: "valid bundle no icon",
-			directory:   "./testdata/valid_bundle_no_icon",
```
camilamacedo86, line 142: we do not add and read a bundle. we can just mock the data to be able to cover all scenarios. 
See that we are now covering all issues/errors that this check can raise instead of only checking only one scenario as before.

pkg/validation/internal/testdata/valid_bundle_with_v1beta1_webhooks/cache.example.com_memcacheds.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: apiextensions.k8s.io/v1
```
camilamacedo86, line 1: added for we cover the api deprecation checks.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +91,186 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	if len(strings.Split(bundle.CSV.Spec.MinKubeVersion, ".")) == 2 {
+		bundle.CSV.Spec.MinKubeVersion += ".0"
+	}
+	semverMinKube, _ := semver.Make(bundle.CSV.Spec.MinKubeVersion)
+
+	// Check if the versions found/informed are that ones which the deprecated APIS are no longer supported (>= 1.22)
+	isUnsupported := semK8s.GE(k8s1_22) || semverMinKube.GE(k8s1_22)
+
+	// If the operator bundle is configured to support all k8s versions or any version >= 16
+	if len(k8sVersionFound) == 0 || semK8s.GE(k8s1_16) {
```
kevinrizza, line 155: Isn't this also true if you just don't pass a value in? Are we making the k8sVersion parameter required?

pkg/validation/internal/operatorhub.go
```diff
@@ -20,48 +20,58 @@ import (
...
+	"OpenShift Optional":     {},
+	"Security":               {},
+	"Storage":                {},
+	"Streaming & Messaging":  {},
 }
 
 func validateOperatorHub(objs ...interface{}) (results []errors.ManifestResult) {
+
+	k8sVersion := ""
+	for _, obj := range objs {
```
kevinrizza, line 57: Why do this here and overload the validate method? I'm wondering if this should just happen further into the validation method that cares about it.

kevinrizza (commented):
This looks really close. Just a couple of small comments mostly around the new parameter.

pkg/validation/internal/operatorhub.go
camilamacedo86: It is inside of the only validator that uses it. See that it is inside of `validateOperatorHub` and will not be done for any other validator. Note that in SDK we will add a flag `--optional-values="k8s=1.22"` to pass optional values. So, if a specific validator requires/uses any optional value we would be able to get that. 

We might want other keys=values in the future.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +91,186 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	semK8s, _ := semver.Make(k8sVersionFound)
 
-	for _, maintainer := range csv.Spec.Maintainers {
-		if maintainer.Name == "" || maintainer.Email == "" {
-			errs = append(errs, fmt.Errorf("csv.Spec.Maintainers elements should contain both name and email"))
+	// Transform the spec minKubeVersion in semver Version to compare
+	if len(strings.Split(bundle.CSV.Spec.MinKubeVersion, ".")) == 2 {
+		bundle.CSV.Spec.MinKubeVersion += ".0"
+	}
+	semverMinKube, _ := semver.Make(bundle.CSV.Spec.MinKubeVersion)
```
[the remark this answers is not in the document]
camilamacedo86: good catcher. I forgot these ones. Done + covered in the tests.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +91,186 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
 
-	if csv.Spec.Provider.Name == "" {
-		errs = append(errs, fmt.Errorf("csv.Spec.Provider.Name not specified"))
+	// If the K8S version was informed via an flag/option then, we should use it
+	k8sVersionFound = k8sVersion
+	if len(strings.Split(k8sVersionFound, ".")) == 2 {
+		// if the value found is like 1.16 then, we concat with .0 to check it with semver
+		k8sVersionFound += ".0"
 	}
+	semK8s, _ := semver.Make(k8sVersionFound)
```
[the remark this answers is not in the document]
camilamacedo86: good catcher. I forgot these ones. Done + covered in the tests.

pkg/validation/internal/operatorhub.go
camilamacedo86: I improved the code here to make that more clear + add some comments. 
- If the option key is informed than we use it
- (So, if that was not informed or it is >= 1.16 we need to do the check)

If the key is not informed we will consider the minKubeVersion. However, if the minKubeVersion is empty for example it means that the operator bundle can be distributed for any k8s version and we should check and raise a warning.  If the minKubeVersion >= 1.22 then, we can safely raise the issue instead.

pkg/validation/internal/operatorhub.go
camilamacedo86: See: (These changes tested/checked with SDK)

Screen Shot 2021-03-24 at 00 20 55

See the help: 

Screen Shot 2021-03-24 at 00 22 26

camilamacedo86: It is inside of the `ValidateOperatorHub` only. The idea here is to allow us to be able to pass a map[string] since for the CLI we need a generic flag option which indeed can be used by any other optional validator. We might want to inform other optional values in the future as well.   

I added how it will be used in SDK. See: https://github.com/operator-framework/api/pull/103#issue-599058459. I hope that clarifies the motivations. 

Please, let me know if it answers your question.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+
+	}
+	return nil
+}
+
+// validateHubDeprecatedAPIS will check if the operator bundle is using a deprecated or no longer supported k8s api
+func validateHubDeprecatedAPIS(bundle *manifests.Bundle, k8sVersion string) ([]error, []error) {
 	var errs []error
 	var warns []error
+	k8s1_22, _ := semver.Make("1.22.0")
```
dinhxuanvu, line 143: Can we comment on what these semver versions mean? Why do these k8s versions matter in the grand scheme of things? I can guess these are upper and lower bounds for k8s version.
Perhaps we should convert those two to const so we update them in the future?

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+// authors knows if their operator bundles are or not respecting the Naming Convention Rules.
+// However, the operator authors still able to choose the names as please them.
+func validateHubChannels(channels []string) error {
+	const preview = "preview"
+	const stable = "stable"
+	const fast = "fast"
+	for _, channel := range channels {
+		if !strings.Contains(channel, preview) &&
+			!strings.Contains(channel, stable) &&
+			!strings.Contains(channel, fast) {
```
estroz, line 130: So package maintainers can only use channel strings that contain one of the above 3? Why can't a maintainer have these 3 plus other arbitrary channel names?

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	return false
+}
+
+// validateHubCSVSpec will check the CSV against the criteria to publish an
+// operator bundle in the OperatorHub.io
+func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) ([]error, []error) {
+	var errs []error
+	var warns []error
+
+	errs = checkSpecProviderName(csv, errs)
```
dinhxuanvu, line 238: I feel these functions are very similar so maybe we should make their return signature identical (return errs and warns). Sure some of them don't have warnings for now but maybe they will in the future? This is just a nit. I don't want to drag this down as it is not important.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+func validateHubChannels(channels []string) error {
+	const preview = "preview"
+	const stable = "stable"
+	const fast = "fast"
+	for _, channel := range channels {
+		if !strings.Contains(channel, preview) &&
+			!strings.Contains(channel, stable) &&
+			!strings.Contains(channel, fast) {
+			return fmt.Errorf("ensure that your channels are following the channel naming convention. " +
+				"More info: https://olm.operatorframework.io/docs/best-practices/channel-naming/")
```
estroz, line 132:
```suggestion
			return fmt.Errorf("channel %q does not follow channel naming conventions: " +
				"https://olm.operatorframework.io/docs/best-practices/channel-naming/")
```

dinhxuanvu (approved):
Just some comments that are nits.
/lgtm

pkg/validation/internal/operatorhub.go
estroz, line 143:
```suggestion
	k8s1_22 := semver.MustParse("1.22.0")
```
Also consider making this a global var, since this parse step occurs for each call to this function.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	}
+	return nil
+}
+
+// validateHubDeprecatedAPIS will check if the operator bundle is using a deprecated or no longer supported k8s api
+func validateHubDeprecatedAPIS(bundle *manifests.Bundle, k8sVersion string) ([]error, []error) {
 	var errs []error
 	var warns []error
+	k8s1_22, _ := semver.Make("1.22.0")
+	k8s1_16, _ := semver.Make("1.16.0")
```
estroz, line 144:
```suggestion
	k8s1_16 := semver.MustParse("1.16.0")
```

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	// Transform the key/option k8sVersion in semver Version to compare
+	var semverK8sVersion semver.Version
+	if hasK8sVersion {
+		version := k8sVersion
+		if len(strings.Split(k8sVersion, ".")) == 2 {
+			// if the value found is like 1.16 then, we concat with .0 to check it with semver
+			version += ".0"
+		}
+		var err error
+		semverK8sVersion, err = semver.Make(version)
```
estroz, line 156:
```suggestion
		var err error
		semverK8sVersion, err = semver.ParseTolerant(k8sVersion)
```

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("invalid value informed via the k8s key option : %s", k8sVersion))
+		} else {
+			// we might want to return it as info instead of warning in the future.
+			warns = append(warns, fmt.Errorf("checking APIs against Kubernetes version : %s", k8sVersion))
```
estroz, line 161: Why would you warn if version parsing was successful?

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+		minKube := bundle.CSV.Spec.MinKubeVersion
+		if len(strings.Split(bundle.CSV.Spec.MinKubeVersion, ".")) == 2 {
+			minKube += ".0"
 		}
-		if maintainer.Email != "" {
-			_, err := mail.ParseAddress(maintainer.Email)
-			if err != nil {
-				errs = append(errs, fmt.Errorf("csv.Spec.Maintainers email %s is invalid: %v", maintainer.Email, err))
+		var err error
+		if semverMinKube, err = semver.Make(minKube); err != nil {
```
estroz, line 173: Use `semver.ParseTolerant()` here too.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+}
+
+// hasDeprecatedAPIs returns true when the bundle has one CRD/Webhook using v1beta1
+func hasDeprecatedAPIs(bundle *manifests.Bundle) bool {
+	if len(bundle.V1beta1CRDs) > 0 {
+		return true
+	}
+	for _, obj := range bundle.CSV.Spec.WebhookDefinitions {
+		for _, v := range obj.AdmissionReviewVersions {
+			if v == "v1beta1" {
```
estroz, line 209: This is the version of admission reviews, not of the webhook config itself which can be either of `v1beta1` or `v1`. Make sure the error message says to check both admission review versions and webhook config version.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	// However, if the minKubeVersion not be informed then, it means that the operator can be distributed in any version.
+	// To ensure this scenario, when the k8sVersion is not informed, the check is made and a warning is raised
+	// if the deprecated/unsupported APIs are found. See that the error only can be safely raised when the
+	// minKubeVersion value is >=1.22.
+	if !hasK8sVersion || semverK8sVersion.GE(k8s1_16) {
+		isUnsupported := semverK8sVersion.GE(k8s1_22) || semverMinKube.GE(k8s1_22)
+
+		// Note that if the operator bundle has both versions then, it will only raise a warning since it might
+		// can means that the operator has been covering both scenarios.
+		if hasDeprecatedAPIs(bundle) {
```
estroz, line 190: I would make these checks and resulting error messages specific to the API type. Getting a list of possible GKs is confusing, because one or multiple of them could exist in a project.

pkg/validation/internal/operatorhub.go
```diff
@@ -17,51 +17,71 @@ import (
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
 )
 
+const minKubeVersionWarnMessage = "csv.Spec.minKubeVersion is not informed. It is recommended you provide this information. " +
+	"Otherwise, it would mean that your operator project can be distributed and installed in any cluster version " +
+	"available, which is not necessarily the case for all projects."
+const apiDeprecationMsg = "apiextensions/v1beta1 for CRDs and admissionregistration.k8s.io/v1beta1 for Webhooks was " +
```
estroz, line 23: Split this into one message per GK.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	errs, warns = checkBundleName(csv, warns, errs)
+
+	return errs, warns
+}
+
+// checkBundleName will validate the operator bundle name informed via CSV.metadata.name.
+func checkBundleName(csv v1alpha1.ClusterServiceVersion, warns []error, errs []error) ([]error, []error) {
+	if len(strings.TrimSpace(csv.Name)) == 0 {
+		errs = append(errs, fmt.Errorf("csv.metadata.Name is empty. You should inform the operator bundle name e.g memcached-operator.v0.0.1"))
+	} else {
```
estroz, line 254: This check is already performed by [the CSV validator](https://github.com/operator-framework/api/blob/master/pkg/validation/internal/csv.go#L35-L36)
```suggestion
	if len(strings.TrimSpace(csv.Name)) != 0 {
```

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+		// The motivation for the following check is to ensure that operators authors knows that operator bundles names should
+		// follow a name convention with versioning
+		bundleNameSplit := strings.Split(csv.Name, ".v")
```
estroz, line 257: Is `foo-operator.0.0.1` not allowed? `0.0.1` is stricter semver than `v0.0.1`.

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+func checkSpecMinKubeVersion(csv v1alpha1.ClusterServiceVersion, warns []error, errs []error) ([]error, []error) {
+	if len(strings.TrimSpace(csv.Spec.MinKubeVersion)) == 0 {
+		warns = append(warns, fmt.Errorf(minKubeVersionWarnMessage))
+	} else {
+		version := csv.Spec.MinKubeVersion
+		if len(strings.Split(csv.Spec.MinKubeVersion, ".")) == 2 {
+			version += ".0"
+		}
+		_, err := semver.Parse(version)
+		if err != nil {
```
estroz, line 281:
```suggestion
		if _, err := semver.ParseTolerant(csv.Spec.MinKubeVersion); err != nil {
```

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+		}
+	}
+	return errs, warns
+}
 
+// checkSpecVersion will validate the spec Version informed via CSV.spec.Version
+func checkSpecVersion(csv v1alpha1.ClusterServiceVersion, errs []error) []error {
+	// spec.Version needs to be set
+	emptyVersion, _ := semver.New("0.0.0")
+	if csv.Spec.Version.Equals(*emptyVersion) {
```
estroz, line 292:
```suggestion
	if csv.Spec.Version.Equals(semver.Version{}) {
```

pkg/validation/internal/operatorhub.go
```diff
@@ -81,41 +101,202 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+		errs = append(errs, fmt.Errorf("csv.Spec.Version is not set"))
```
estroz, line 293:
```suggestion
		errs = append(errs, fmt.Errorf("csv.Spec.Version must be set"))
```

pkg/validation/internal/operatorhub.go
```diff
@@ -126,12 +307,37 @@ func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) ([]error, []error) {
...
+		// use custom categories for validation if provided
+		customCategoriesPath := os.Getenv("OPERATOR_BUNDLE_CATEGORIES")
+		if customCategoriesPath != "" {
+			customCategories, err := extractCategories(customCategoriesPath)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("could not extract custom categories from categories %#v: %s", customCategories, err))
+			} else {
+				for _, category := range categorySlice {
+					if _, ok := customCategories[strings.TrimSpace(category)]; !ok {
+						errs = append(errs, fmt.Errorf("csv.Metadata.Annotations.Categories %s is not a valid custom category", category))
```
estroz, line 322:
```suggestion
						errs = append(errs, fmt.Errorf("csv.Metadata.Annotations[\"categories\"] value %s is not in the set of custom categories", category))
```

pkg/validation/internal/operatorhub.go
```diff
@@ -126,12 +307,37 @@ func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) ([]error, []error) {
...
+					}
+				}
+			}
+		} else {
+			// use default categories
+			for _, category := range categorySlice {
+				if _, ok := validCategories[strings.TrimSpace(category)]; !ok {
+					errs = append(errs, fmt.Errorf("csv.Metadata.Annotations.Categories %s is not a valid category", category))
```
estroz, line 330:
```suggestion
					append(errs, fmt.Errorf("csv.Metadata.Annotations[\"categories\"] value %s is not in the set of default categories", category))
```

pkg/validation/internal/testdata/valid_bundle_with_only_v1beta1_webhooks/memcached-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,247 @@
...
+      name: memcacheds.cache.example.com
+      version: v1alpha1
+  description: Memcached Operator description. TODO.
+  displayName: Memcached Operator
+  icon:
+  - base64data: ""
+    mediatype: ""
+  install:
+    spec:
+      clusterPermissions:
```
estroz, line 37: Nit: to reduce the amount of testdata, remove extraneous config like permissions/deployments (or at least use minimal values).

estroz (changes requested):

pkg/validation/internal/operatorhub.go
camilamacedo86: @estroz it is clarified in the comment of this check. These are the naming suggested "convention". See that will raise a warning and will share the link: https://olm.operatorframework.io/docs/best-practices/channel-naming/ to ensure that they know about that. However, they are still able to choose the names which please them.

pkg/validation/internal/operatorhub.go
camilamacedo86: See: https://github.com/operator-framework/api/blob/00a685b5012bc776d277eb07e77245b33ff21e16/pkg/validation/internal/operatorhub.go#L179-L184

pkg/validation/internal/operatorhub.go
estroz: Right, but I should be able to define any channel name I want as long as preview/fast/stable are also defined for a bundle. This check really should be
```go
for _, conventionalChannelPrefix := range []string{preview, fast, stable} {
	if !containsPrefix(channels, conventionalChannelPrefix) {
		errs = append(errs, fmt.Errorf("bundle channels should contain a %s channel", conventionalChannel))
	}
}
```

pkg/validation/internal/operatorhub.go
estroz: That isn't true. This code adds this warning regardless of whether deprecated APIs exist.

pkg/validation/internal/operatorhub.go
camilamacedo86: Closing that as solved. However, feel free to re-open if you see that is required.

pkg/validation/internal/operatorhub.go
camilamacedo86: Closing that as solved. However, feel free to re-open if you see that is required.

pkg/validation/internal/operatorhub.go
camilamacedo86: Ok for the suggestion. However, what do you mean with make global var? It is not used outside of this method.

pkg/validation/internal/operatorhub.go
camilamacedo86: hi @estroz, 

> This code adds this warning regardless of whether deprecated APIs exist.
Yes, we should NOT return an error in all scenarios. Please, see the comment above. It describes when the error should not be raised and why.

pkg/validation/internal/operatorhub.go
camilamacedo86: done

pkg/validation/internal/operatorhub.go
camilamacedo86: done

pkg/validation/internal/operatorhub.go
camilamacedo86: I think it is done. So, I am closing this one. However, feel free to re-open if you need.

pkg/validation/internal/operatorhub.go
camilamacedo86: I improved the explanation/comments. Also, might easier to check the tests to get better the rules.
So, I am closing this one. However, please feel free to re-open if required.

pkg/validation/internal/testdata/valid_bundle_with_only_v1beta1_webhooks/memcached-operator.clusterserviceversion.yaml
camilamacedo86: The test data scenarios here are re-used for more than one check. So, I think is better we leave that since it might be useful for another case.

pkg/validation/internal/operatorhub.go
estroz: A warning shouldn’t be returned either, since nothing is wrong with user inputs. How about just logging an info message?

camilamacedo86: - The error/warning is returned when the operator bundle has any manifest using the deprecated APIs and the tests are considering a version >= 1.16
- However, the error is ONLY returned if the k8s flag informed or minKubeVersion is >= 1.22 and the operator bundles do NOT have any manifest with the latest api version. 

See the scenarios; 

https://github.com/operator-framework/api/blob/facca9fbef4d53e06a83e41e2183c6854bec56b3/pkg/validation/internal/operatorhub_test.go#L325-L462

Then, what is the scenario that in POV is missing here?

camilamacedo86: @estroz thank you for your review. I think that is right now. 
also, checking further in OperatorHub will not possible either publish a project valid for both scenarios. (OLM will try to install all. ) so, it makes the tests/checks simpler as well.

pkg/validation/internal/testdata/valid_bundle_with_only_v1beta1_webhooks/memcached-operator.clusterserviceversion.yaml
estroz: If those tests need permissions etc., then they can be added later. Right now they don't. However its just a nit so feel free to ignore.

pkg/validation/internal/operatorhub.go
estroz: >The error/warning is returned when the operator bundle has any manifest using the deprecated APIs and the tests are considering a version >= 1.16 (we just check if has or not deprecated apis)

I'm saying that this is not true. The following summary of the current validator implementation will always return an error if `versionProvided != ""`, which effectively means that a warning will be emitted any time a user specifies a k8s version external to the bundle's `minKubeVersion`, regardless of the version specified.

```go
func validateHubDeprecatedAPIS(bundle *manifests.Bundle, versionProvided string) ([]error, []error) {
	isVersionProvided := len(versionProvided) > 0

	...

	if isVersionProvided {
		semVerVersionProvided, err = semver.ParseTolerant(versionProvided)
		if err != nil {
			errs = append(errs, fmt.Errorf("invalid value informed via the k8s key option : %s", versionProvided))
		} else {
			// we might want to return it as info instead of warning in the future.
			warns = append(warns, fmt.Errorf("checking APIs against Kubernetes version : %s", versionProvided))
		}
	}
...
}
```

Is that your intent? The tests seem to say so, and I don't think a warning should be emitted here. What if a validator consumer, like operator-sdk, were to add an option like `--fail-if-warn`? Then this validator would fail erroneously.

camilamacedo86: Then, it means that:

--optional-values="k8s=value" flag with a value => 1.16 <= 1.22 the validator will return result as warning.
--optional-values="k8s=value" flag with a value => 1.22 the validator will return result as error.
minKubeVersion >= 1.22 return the error result.


Yes. If minKubeVersion is Empty then it means that we should raise a WARN. It is the same as 1.11+

pkg/validation/internal/operatorhub.go
camilamacedo86: solved we will be unable to check that
and actually it might not be important as well such as we spoke

pkg/validation/internal/operatorhub.go
camilamacedo86: done

pkg/validation/internal/operatorhub.go
camilamacedo86: I think it is no longer required since we check the crds and raise the info msg.
So, I am closing. However, please feel free to re-open if you see that we should change that.

pkg/validation/internal/operatorhub.go
camilamacedo86: All unique value is allowed. We just raise the WARN here when it is not following the convention which is with `name.v<semver>`. Do you think that we should check if is `foo-operator.0.0.1` and not raise the WARN in this case?

pkg/validation/internal/operatorhub.go
estroz: In that case, remove the webhook bit from the message

pkg/validation/internal/operatorhub.go
```diff
@@ -17,51 +17,72 @@ import (
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
 )
 
+const minKubeVersionWarnMessage = "csv.Spec.minKubeVersion is not informed. It is recommended you provide this information. " +
+	"Otherwise, it would mean that your operator project can be distributed and installed in any cluster version " +
+	"available, which is not necessarily the case for all projects."
+
+const apiDeprecationMsg = "apiextensions/v1beta1 for CRDs and admissionregistration.k8s.io/v1beta1 for Webhooks was " +
+	"deprecated in Kubernetes `1.16` and removed in `1.22`."
```
estroz, line 25:
```suggestion
const crdv1beta1DeprecationMsg = "apiextensions.k8s.io/v1beta1, kind=CustomResourceDefinitions was deprecated in Kubernetes v1.16 and will be removed in 1.22 in favor of v1"
```

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +95,301 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+			!strings.Contains(channel, stable) &&
+			!strings.Contains(channel, fast) {
+			channelsNotFollowingConventional = append(channelsNotFollowingConventional, channel)
+		}
+
+	}
+
+	if len(channelsNotFollowingConventional) > 0 {
+		return fmt.Errorf("channel(s) %q are not following the recommended naming convention:"+
+			"https://olm.operatorframework.io/docs/best-practices/channel-naming/", "'"+strings.Join(channelsNotFollowingConventional, `','`)+`'`)
```
estroz, line 141: The `%+q` formatting element will wrap a list of strings in a nicely quoted list-y string.
```suggestion
		return fmt.Errorf("channel(s) %+q are not following the recommended naming convention: "+
			"https://olm.operatorframework.io/docs/best-practices/channel-naming",
			channelsNotFollowingConventional)
```

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +95,301 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	}
+
+	return nil
+}
+
+func containsPrefix() {
```
estroz, line 147: ?

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +95,301 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+// the CSV to do the checks. So, the criteria is >=minKubeVersion. By last, if the minKubeVersion is not provided
+// then, we should consider the operator bundle is intend to work well in any Kubernetes version.
+// Then, it means that:
+//--optional-values="k8s=value" flag with a value => 1.16 <= 1.22 the validator will return result as warning.
+//--optional-values="k8s=value" flag with a value => 1.22 the validator will return result as error.
+//minKubeVersion >= 1.22 return the error result.
+//minKubeVersion empty returns a warning since it would mean the same of allow install in any supported version
+func validateHubDeprecatedAPIS(bundle *manifests.Bundle, versionProvided string) ([]error, []error) {
 	var errs []error
 	var warns []error
```
estroz, line 162: Nit
```suggestion
func validateHubDeprecatedAPIS(bundle *manifests.Bundle, versionProvided string) (errs, warns []error) {
```

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +95,301 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+			}
+
+			// isUnsupported is true only if the key/value OR minKubeVersion were informed and are >= 1.22
+			isUnsupported := semVerVersionProvided.GE(semVerK8sVerV1betav1Unsupported) ||
+				semverMinKube.GE(semVerK8sVerV1betav1Unsupported)
+			// We only raise an error when the version >= 1.22 was informed via
+			// the k8s key/value option or is specifically defined in the CSV
+			if isUnsupported {
+				errs = append(errs, fmt.Errorf("%s Use the version v1 instead for the APIs : %s",
+					apiDeprecationMsg, "'"+strings.Join(crdApiNames, `','`)+`'`))
```
estroz, line 218:
```suggestion
				errs = append(errs, fmt.Errorf("%s: %+q", apiDeprecationMsg, crdApiNames))
```

pkg/validation/internal/operatorhub.go
camilamacedo86: cool

pkg/validation/internal/operatorhub.go
camilamacedo86: Any unique value is allowed. We just WARN the scenarios. 
If it has not 0.0.1 -> then is not following the versioning convention
if it is not operator-name.v<semver> -> is not following the naming convention

I improved that using regex as should be. Thank you.

camilamacedo86:
Hi @estroz, 

thank you for your great review. 
I think it is shaped enough to get merged now. WDYT?
Do you see any other thing that we should improve here?

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +96,303 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	warns []error
+}
+
+// checkBundleName will validate the operator bundle name informed via CSV.metadata.name.
+// The motivation for the following check is to ensure that operators authors knows that operator bundles names should
+// follow a name and versioning convention
+func checkBundleName(checks CSVChecks) CSVChecks {
+
+	// Check if is following the semver
+	re := regexp.MustCompile("([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z-]+)?$")
```
estroz, line 251: This regexp is incorrect since it allows multiple hyphens and weird garbage at the end of the version itself. How about splitting the string by `.` and parsing the last n-1 values with the semver lib after stripping “v” from the first of them? That should always give you a correct value and avoids complex regexp.

pkg/validation/internal/operatorhub.go
```diff
@@ -17,51 +18,72 @@ import (
...
 }
 
 func validateOperatorHub(objs ...interface{}) (results []errors.ManifestResult) {
+
+	// Obtain the k8s version if informed via the objects an optional
+	k8sVersion := ""
+	for _, obj := range objs {
+		switch obj.(type) {
+		case map[string]string:
+			k8sVersion = obj.(map[string]string)["k8s"]
```
estroz, line 69: It would be nice to define something like terraform’s [`TF_VARS_name`](https://www.terraform.io/docs/cli/config/environment-variables.html) for validators, where the prefix is the validator name capitalized, so you don’t have to pass in some object or define flags for the validator lib consumer.

Just an idea for a follow up, this is fine for now.

pkg/validation/internal/operatorhub.go
camilamacedo86: Its goal is to get what the authors added as a version. I added a test to prove that the warning will work well if the name has multiple hyphens as described. See (memcached-operator.v1--1.0). Then, I would prefer to move forward here with the regex it shows easier to keep maintained and more assertive.

pkg/validation/internal/operatorhub.go
camilamacedo86: I am not sure about that. Why should we adopt a Terraform convention?

pkg/validation/internal/operatorhub.go
estroz: If the "extra stuff" formatting is intentional, then I'll resolve this issue.

pkg/validation/internal/operatorhub.go
estroz: My suggestion is not a terraform convention, just an idea lifted from terraform.

pkg/validation/internal/operatorhub.go
```diff
@@ -74,111 +96,303 @@ func validateBundleOperatorHub(bundle *manifests.Bundle) errors.ManifestResult {
...
+	// Transform the key/option versionProvided in semver Version to compare
+	var semVerVersionProvided semver.Version
+	if isVersionProvided {
+		var err error
+		semVerVersionProvided, err = semver.ParseTolerant(versionProvided)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("invalid value informed via the k8s key option : %s", versionProvided))
+		} else {
+			// we might want to return it as info instead of warning in the future.
+			warns = append(warns, fmt.Errorf("checking APIs against Kubernetes version : %s", versionProvided))
```
estroz, line 179: I still have a negative opinion about warning here. However I won't let it block the PR. This can always be changed later.

estroz (approved):
/lgtm

pkg/validation/internal/operatorhub.go
camilamacedo86: IHMO I do not think that would be good to have a prefix here. 
The idea is we are able to use the same key for 1 or Many validators since its usage via SDK can be done via the suite.

PR #104, opened 2021-03-23
scorecard: add `ImagePullSecrets` to the v1alpha3 scorecard config

estroz:
This PR improves the scorecard v1alpha3 config by permitting config owners to specify a secret name for private scorecard images. This is particularly useful for custom image users.

To support arbitrary cluster usage, where the secret name used may differ, a parallel change should be made to scorecard that permits overriding the `name` value of `corev1.LocalObjectReference`.

/kind feature


estroz:
An alternative suggestion, that would require no change in scorecard or the config, is applying the image pull secret [to the service account](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) the test Pods are run under.

estroz:
I'm going to close this for now because the service account approach works and requires no modification.

PR #105, opened 2021-03-29
Revert cleanup spec and status fields

hasbro17:
Reverts https://github.com/operator-framework/api/pull/100

The `spec.cleanup` and `status.cleanup` API fields are no longer needed since the [implementation](https://github.com/operator-framework/operator-lifecycle-manager/pull/2030) for the server-side operand deletion work is on hold and unlikely to proceed per the original enhancement proposal.

See: https://github.com/operator-framework/enhancements/pull/71/files


dinhxuanvu (approved):
/lgtm

hasbro17:
Some concerns were raised over whether this is or isn't a breaking change. It technically is a breaking change that we would normally deprecate first. But given that there wasn't an implementation to accompany this API perhaps removing it outright would be easier.

dinhxuanvu:
Is this cleanup field required? It doesn't appear so to me so it is not much of a breaking change as I thought it would be.

anik120:
Closing this since it's stale now

PR #108, opened 2021-04-08
Add .status.message and .status.startTime to InstallPlan.

benluddy:
StartTime is intended to be set by the controller when plan execution
begins (that is, after approval and bundle unpacking, when step
resources may actually be applied to the cluster) and supports retry
and timeout features.

Message provides a field to communicate detailed free-form messages to
administrators about the status of the plan.


hasbro17 (approved):
/lgtm

PR #110, opened 2021-04-14
Add UID Optional field to scorecard config types

theishshah:
Create optional Unique ID field for Scorecard config types.


pkg/apis/scorecard/v1alpha3/configuration_types.go
```diff
@@ -34,6 +34,8 @@ type StageConfiguration struct {
 type TestConfiguration struct {
 	// Image is the name of the test image.
 	Image string `json:"image" yaml:"image"`
+	// UniqueID is is an optional unique test identifier of the test image.
```
estroz, line 37: Can you add an example of a default/generated value used if this is unset, and clarify where the default/generated value comes from?

estroz (changes requested):

jmrodri (approved):
/lgtm

Assuming @estroz concern is addressed.

pkg/apis/scorecard/v1alpha3/configuration_types.go
theishshah: We are not going default this value, as it's not being used anywhere within scorecard and is only available as an option if someone wants to specifically tag a test and watch it with a unique id for that run.

jmrodri (approved):
/lgtm

pkg/apis/scorecard/v1alpha3/configuration_types.go
estroz: Does this value conform to any specifications, ex. DNS 1123? What do you mean by "watch it"? As a user, how and where is this value applied to my scorecard tests? Are there examples of what value I might set here?

These are some of the questions not answered by any documentation/godocs anywhere, which is kinda what I was getting at with my question. It isn't clear what this field does, and libraries should be self-descriptive.

PR #113, opened 2021-04-23
operators/v1alpha1: add ConversionReviewversions and GetConversionWebhook()

estroz:
This PR adds the ConversionWebhookDescription type and native webhook type-building methods to `pkg/operators/v1alpha1` to handle conversion webhooks properly and differentiate from admission webhooks.

/kind feature


pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -176,18 +177,25 @@ type WebhookDescription struct {
...
-	SideEffects             *admissionregistrationv1.SideEffectClass        `json:"sideEffects"`
-	TimeoutSeconds          *int32                                          `json:"timeoutSeconds,omitempty"`
-	AdmissionReviewVersions []string                                        `json:"admissionReviewVersions"`
+	ContainerPort  int32                                        `json:"containerPort,omitempty"`
+	TargetPort     *intstr.IntOrString                          `json:"targetPort,omitempty"`
+	Rules          []admissionregistrationv1.RuleWithOperations `json:"rules,omitempty"`
+	FailurePolicy  *admissionregistrationv1.FailurePolicyType   `json:"failurePolicy,omitempty"`
+	MatchPolicy    *admissionregistrationv1.MatchPolicyType     `json:"matchPolicy,omitempty"`
+	ObjectSelector *metav1.LabelSelector                        `json:"objectSelector,omitempty"`
+	SideEffects    *admissionregistrationv1.SideEffectClass     `json:"sideEffects,omitempty"`
```
estroz, line 186: Note: added `omitempty` here since conversion webhooks don't need these.

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -176,18 +177,25 @@ type WebhookDescription struct {
...
+	TimeoutSeconds *int32                                       `json:"timeoutSeconds,omitempty"`
+	// AdmissionReviewVersions is set to one or more versions of the admission.k8s.io AdmissionReview API.
+	// See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response for available versions.
+	// These versions are only valid if Type is ValidatingAdmissionWebhook, MutatingAdmissionWebhook.
+	AdmissionReviewVersions []string                                        `json:"admissionReviewVersions,omitempty"`
```
estroz, line 191: Note: added `omitempty` here since conversion webhooks don't need these.

pkg/operators/v1alpha1/clusterserviceversion_types.go
varshaprasad96: I think ports and `sideEffects` is required for webhookDescription. I remember installplan erroring when one of them were not present. Will verify and update here

estroz: It may be worthwhile to consider having a separate conversion webhook struct in that case, and add it like `.spec.conversionwebhookdefinitions`.

varshaprasad96: Yes, it would be helpful to have a different `conversionwebhookdefinition` struct. Because port and `conversionReviewVersions` are required for conversion WH and not for others

dinhxuanvu:
Not sure if v1alpha1 is the correct place for it. You did add a new blob to CSV in v1alpha1 so maybe it should be in v1alpha2?

PR #114, opened 2021-04-27
Add CatSrc conditions for err during registry pod creation workflow

anik120:


anik120:
Closed https://github.com/openshift/operator-framework-olm/pull/66
But keeping this open for now since we do need to address this https://github.com/operator-framework/operator-lifecycle-manager/blob/d980191ecdb848c6538b721906dfc993e86edd6f/pkg/controller/registry/reconciler/grpc.go#L195 at some point.

PR #119, opened 2021-05-10
Add conditions array to OperatorCondition's spec

dinhxuanvu:
The conditions array in the spec is now available for operator to
create/update as the operator progresses through installation
process. As the spec is updated, the object generation will be
incremented and it can be used for tracking object changes.


dinhxuanvu:
The proposal: https://hackmd.io/9wG20hu5TU-y1HrkhvcsZQ?view FYI

pkg/operators/v2/operatorcondition_types.go
```diff
@@ -0,0 +1,51 @@
...
+
+// OperatorConditionSpec allows a cluster admin to convey information about the state of an operator to OLM, potentially overriding state reported by the operator.
+type OperatorConditionSpec struct {
+	ServiceAccounts []string           `json:"serviceAccounts,omitempty"`
+	Deployments     []string           `json:"deployments,omitempty"`
+	Overrides       []metav1.Condition `json:"overrides,omitempty"`
+	Conditions      []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// OperatorConditionStatus allows an operator to convey information its state to OLM. The status may trail the actual
```
awgreene, line 20:
```suggestion
// OperatorConditionStatus allows OLM to convey which conditions have been observed.
```

pkg/operators/v2/operatorcondition_types.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Upgradeable indicates that the operator is upgradeable
+	Upgradeable string = "Upgradeable"
+)
+
+// OperatorConditionSpec allows a cluster admin to convey information about the state of an operator to OLM, potentially overriding state reported by the operator.
```
awgreene, line 12:
```suggestion
// OperatorConditionSpec allows an operator to report state to OLM and provides cluster admin with the ability to manually override state reported by the opeator.
```

awgreene (changes requested):
Great turn around time on this @dinhxuanvu, left some nits.

pkg/operators/v2/operatorcondition_types.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// OperatorConditionStatus allows OLM to convey which conditions have been observed.
+type OperatorConditionStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
+// +kubebuilder:storageversion
```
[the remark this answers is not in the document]
dinhxuanvu: Sure.

pkg/operators/v2/operatorcondition_types.go
```diff
@@ -0,0 +1,51 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Upgradeable indicates that the operator is upgradeable
+	Upgradeable string = "Upgradeable"
+)
+
+// OperatorConditionSpec allows an operator to report state to OLM and provides
+// cluster admin with the ability to manually override state reported by the operator.
```
awgreene, line 13: admins
```suggestion
// cluster admins with the ability to manually override state reported by the operator.
```

awgreene (changes requested):
2 small nits

pkg/operators/v2/operatorcondition_types.go
awgreene: Sorry - this was my suggested mistake haha

awgreene (approved):

estroz:
Out of curiosity: why v2 and not v2beta1? Are we confident no significant changes will be needed soon? I feel that it’s a bit hasty to just throw a new stable version out, especially given how short the lifetime of v1 was.

awgreene:
>Out of curiosity: why v2 and not v2beta1? Are we confident no significant changes will be needed soon? I feel that it’s a bit hasty to just throw a new stable version out, especially given how short the lifetime of v1 was.

Spoke with Eric about this offline, we could have simply added the new fields to v1 but OLM would be unable to identify users that are still writing to the broken v1 API. Additionally, there were concerns that users would rely on a stable (v1)version over a beta/alpha version.

PR #127, opened 2021-06-04
Introduce verify github action

awgreene:
Problem: It is possible to commit changes to the repo without
updating generated files.

Solution: Introduce a GitHub Action which ensures that generated
files are up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ install: ## Build & install operator-verify
 		$(BUILD_PATH)
 
 # Code management.
-.PHONY: format tidy clean vendor generate
+.PHONY: format tidy clean vendor generate manifests
 
 format: ## Format the source code
 	$(Q)go fmt $(PKGS)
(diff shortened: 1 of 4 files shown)


dinhxuanvu:
Please discard the changes on `zz_generated.deepcopy.go`. That's not relevant to this PR.

awgreene:
> Please discard the changes on `zz_generated.deepcopy.go`. That's not relevant to this PR.

I separated the deepcopy changes into this PR: https://github.com/operator-framework/api/pull/128

Makefile
```diff
@@ -77,14 +77,22 @@ manifests: controller-gen ## Generate manifests e.g. CRD, RBAC etc
...
 
 test: test-unit ## Run the tests
 
 TEST_PKGS:=$(shell go list ./...)
 test-unit: ## Run the unit tests
 	$(Q)go test -count=1 -short ${TEST_PKGS}
 
+
+verify-manifests: manifests diff
+verify-generate: generate diff
```
benluddy, line 93: I can't find any promise that GNU Make will build prerequisites in any particular relative order (other than before the target, of course).

timflannagan: You could also just have the `verify` target be a driver that calls the verify-* commands in order instead of supplying them as recipes/dependencies.

awgreene: I was unable to find a guarantee but running the following test always resulted in the expected order:
```
makefile-test: mftest1 mftest2
mftest1:
	echo "1"

mftest2:
	echo "2"
```

awgreene: @benluddy pointed out that the prerequisites for the target above can be executed if the number of jobs is greater than 1.

Example command:
```
$ make makefile-test -j8
echo "1"
echo "2"
2
1
```

PR #134, opened 2021-06-23
Add validation to CSV name length

dinhxuanvu:
Given CSV's name is used as label for OLM, its length should be
validated to ensure it's not over the label character limit
which is currently set at 63 character.

This commit will enable CSV validation returning an error if CSV's
name is more than 63 characters.

### DIFF
--- a/pkg/validation/internal/csv.go
+++ b/pkg/validation/internal/csv.go
@@ -32,11 +32,11 @@ func validateCSVs(objs ...interface{}) (results []errors.ManifestResult) {
 func validateCSV(csv *v1alpha1.ClusterServiceVersion) errors.ManifestResult {
 	result := errors.ManifestResult{Name: csv.GetName()}
 	// Ensure CSV names are of the correct format.
-	if err := parseCSVNameFormat(csv.GetName()); err != (errors.Error{}) {
+	if err := parseCSVNameFormat(csv.GetName()); err != nil {
 		result.Add(errors.ErrInvalidCSV(fmt.Sprintf("metadata.name %s", err), csv.GetName()))
 	}
 	if replaces := csv.Spec.Replaces; replaces != "" {
-		if err := parseCSVNameFormat(replaces); err != (errors.Error{}) {
+		if err := parseCSVNameFormat(replaces); err != nil {
 			result.Add(errors.ErrInvalidCSV(fmt.Sprintf("spec.replaces %s", err), csv.GetName()))
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/validation/internal/testdata/badName.csv.yaml
```diff
@@ -0,0 +1,292 @@
+#! validate-crd: deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
+#! parse-kind: ClusterServiceVersion
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: someoperatorwithanextremelylongnamethatmakenosensewhatsoever.v999.999.999
```
estroz, line 6: Shouldn't this file just contain the necessary fields, ex. just GVK and `metadata.name`?

dinhxuanvu: You want me to slim this down to just basic stuff. Yah, I can do that.

dinhxuanvu: I stripped it down as much as I could. This CSV still goes through the typical validation for the whole CSV so a few things still need to be there.

pkg/validation/internal/csv.go
```diff
@@ -55,6 +55,9 @@ func parseCSVNameFormat(name string) error {
 	if violations := k8svalidation.IsDNS1123Subdomain(name); len(violations) != 0 {
 		return fmt.Errorf("%q is invalid:\n%s", name, violations)
 	}
+	if len(name) > k8svalidation.LabelValueMaxLength {
```
estroz, line 58: If a CSV's name is supposed to be a label, why not just use [`IsDNS1123Label(name)`](https://github.com/kubernetes/apimachinery/blob/v0.21.2/pkg/util/validation/validation.go#L187)?

dinhxuanvu: It is used as label value so I think `IsValidLabelValue` is better. Fixed.

pkg/validation/internal/csv.go
```diff
@@ -55,6 +55,10 @@ func parseCSVNameFormat(name string) error {
 	if violations := k8svalidation.IsDNS1123Subdomain(name); len(violations) != 0 {
 		return fmt.Errorf("%q is invalid:\n%s", name, violations)
 	}
+	// Give CSV name is used as label value, it should be validated
+	if errs := k8svalidation.IsValidLabelValue(name); errs != nil {
+		return fmt.Errorf("%q is invalid: %s", name, strings.Join(errs[:], ","))
+	}
```
estroz, line 61:
```suggestion
	var errStrs []string
	errStrs = append(errStrs, k8svalidation.IsDNS1123Subdomain(name)...)
	// Give CSV name is used as label value, it should be validated
	errStrs = append(errStrs, k8svalidation.IsValidLabelValue(name)...)
	
	if len(errStrs) > 0 {
		return fmt.Errorf("%q is invalid:\n%s", name, strings.Join(errStrs, ","))
	}
	return nil
```

timflannagan: You could also just return a `return utilerrors.NewAggregate(errs)` as that package is already present as a vendored dependency, instead of explicitly checking the length of any list of errors and aggregating an error string.

estroz: The returned type from those validation functions is `[]string` not `[]error`, so you cannot use `NewAggregate()` directly.

timflannagan: Ah, gotcha yeah I should've checked the return type there - good catch. I guess there's no immediate way to wrap a list of strings into a single error to make that type work, so disregard my suggestion.

estroz (approved):
/approve
/lgtm

PR #137, opened 2021-07-07
Add Timestamp to Scorecard Test Results

theishshah:
Add field for time stamp in scorecard results


pkg/apis/scorecard/v1alpha3/test_types.go
```diff
@@ -40,6 +40,7 @@ type TestStatus struct {
 // Test specifies a single test run.
 type Test struct {
 	metav1.TypeMeta `json:",inline"`
+	Tstamp          string            `json:"tstamp,omitempty"`
```
joelanford, line 43: Is it possible to use [`metav1.Time`](https://github.com/kubernetes/apimachinery/blob/8303750571678b1e7ccef2b8b597b5286aab7eb7/pkg/apis/meta/v1/time.go#L24-L33) instead of a string?

joelanford (commented):
Is there some context for the addition of this field that explains what it will be used for?

A few questions in the meantime:
1. What is this the timestamp of? When the test was launched, when it completed, something else?
2. Depending on the answer to 1), I'm wondering if it makes more sense to be in the `TestStatus` or `TestResult` struct?

pkg/apis/scorecard/v1alpha3/test_types.go
theishshah: We can use a time object and convert to a string later when it's needed. I wanted to add this field because Xunit output's would like a timestamp on the test result.
1) Current plan is to time stamp upon the creation of the result
2) Seems reasonable to move it to the result struct instead

theishshah:
Updated to move timestamp to the result struct and changed type to metav1.Time.

pkg/apis/scorecard/v1alpha3/test_types.go
```diff
@@ -28,6 +28,8 @@ type TestResult struct {
 	Errors []string `json:"errors,omitempty"`
 	// Suggestions is a list of suggestions for the user to improve their score (if applicable)
 	Suggestions []string `json:"suggestions,omitempty"`
+	// Timestamp of the result from an invidual test
+	Tstamp metav1.Time `json:"tstamp,omitempty"`
```
joelanford, line 32: Since you mentioned the timestamp would be when the result is created, wdyt about a more descriptive name?

Also GoDoc comments should always start with the name of the identifier it is describing.
```suggestion
	// CreationTimestamp of the result from an invidual test
	CreationTimestamp metav1.Time `json:"creationTimestamp,omitempty"`
```

theishshah: That's fine with me, the x-unit field is assigned from this value but doesn't care what it's called, this more descriptive name will keep the API cleaner going forward.

pkg/apis/scorecard/v1alpha3/test_types.go
```diff
@@ -28,6 +28,8 @@ type TestResult struct {
 	Errors []string `json:"errors,omitempty"`
 	// Suggestions is a list of suggestions for the user to improve their score (if applicable)
 	Suggestions []string `json:"suggestions,omitempty"`
+	// CreationTimestamp of the result from an invidual test
```
joelanford, line 31:
```suggestion
	// CreationTimestamp of the result from an individual test
```

joelanford (commented):
/approve

/hold
In case you want to fix the typo. If you need this to merge ASAP, feel free to just cancel the hold.

PR #139, opened 2021-07-19
Add conditions to catalog source

jchunkins:
This PR adds an optional array of `k8s.io/apimachinery/pkg/apis/meta/v1`.`Condition` whose name is `conditions` to the CatalogSource status. This is being done to allow for condition support for use with the `automatic-catalog-switching-annotations` enhancement proposal [PR](https://github.com/operator-framework/enhancements/pull/87) (which has yet to be merged). 

Note that the initial use of the `conditions` field will be the catalog template resolution logic, and will be used to indicate the status of catalog template resolution. In the future the OLM catalog operator could migrate to using `conditions` instead of the existing `Message` and `Reason` fields but that is out of scope for the enhancement proposal.


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -131,6 +131,16 @@ type CatalogSourceStatus struct {
...
+
+	// Represents the state of a CatalogSource. Note that Message and Reason represent the original
+	// status information, which may be migrated to be conditions based in the future. Any new features
+	// introduced will use conditions.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
```
ecordell, line 143: nit: CRs don't get encoded as protobuf so that tag could be omitted.

ecordell (approved):
/lgtm

timflannagan:
It looks like tide's blocking on an optional github action. I opened https://github.com/openshift/release/pull/20513 to update the tide configuration to alleviate this problem.

PR #142, opened 2021-07-27
Handles incorrect CatalogSource interval values.

DhritiShikhar:
**Description of the change:**

Handles incorrect CatalogSource interval values.

Example:

```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: ibm-operator-catalog
  namespace: openshift-marketplace
spec:
  displayName: ibm-operator-catalog 
  publisher: IBM Content
  sourceType: grpc
  image: docker.io/ibmcom/ibm-operator-catalog
  updateStrategy:
    registryPoll:
      interval: 45mError code    ################### incorrect value
```


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -242,6 +243,28 @@ func (c *CatalogSource) Poll() bool {
...
+		return err
+	}
+	registryPoll := &RegistryPoll{}
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
+	if err != nil {
+		err = nil
+		defaultTime, err := time.ParseDuration("15m")
+		if err != nil {
+			return err
+		}
```
anik120, line 259: nit: 
```suggestion
	if err != nil {
		defaultTime, parseErr := time.ParseDuration("15m")
		if parseErr != nil {
		    return parseErr
		}
```

anik120: Also, the `15m` default time should be initialized as a constant and the constant should be used here instead of hard coding the value here.

anik120 (changes requested):
Thanks for the PR @DhritiShikhar 🎉 
Had a couple of suggestions.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -242,6 +243,28 @@ func (c *CatalogSource) Poll() bool {
 	return true
 }
 
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
```
timflannagan, line 247: It's not immediately clear to me why we'd need to build up this kind of validation check - it's possible I'm missing how we'd use this in OLM. Would it be possible to take a CatalogSource as input, and validate that the `.spec.RegistryPoll.Interval` matches the type signature and format we're expecting? Is this a candidate to enforcing this fields' schema validation such that only time values are acceptable? I'm not sure whether the latter would be a breaking change.

DhritiShikhar: Hi Tim,

This PR is meant to solve this issue -> https://bugzilla.redhat.com/show_bug.cgi?id=1943937

> It's not immediately clear to me why we'd need to build up this kind of validation check

We need to build up this validation check because the operator cannot handle invalid interval values.

We see error in the operator logs:
```
reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1alpha1.CatalogSource: v1alpha1.CatalogSourceList.Items: []v1alpha1.CatalogSource: v1alpha1.CatalogSource.v1alpha1.CatalogSource.Spec: v1alpha1.CatalogSourceSpec.UpdateStrategy: v1alpha1.UpdateStrategy.RegistryPoll: v1alpha1.RegistryPoll.Interval: unmarshalerDecoder: time: unknown unit "mCopy code" in duration "45mError code", error found in #10 byte of ...|Copy code"}}}},{"api|..., bigger context ...|rategy":{"registryPoll":{"interval":"45mCopy code"}}}},{"apiVersion":"operators.coreos.com/v1alpha1"|...
```

> Would it be possible to take a CatalogSource as input, and validate that the .spec.RegistryPoll.Interval matches the type signature and format we're expecting

Is there an advantage to this approach?

> Is this a candidate to enforcing this fields' schema validation such that only time values are acceptable?

The user can still put any value such as:

```
  updateStrategy:
    registryPoll:
      interval: 45mError code    ################### incorrect value
```

But since it is invalid, it would get converted to default interval value.

> I'm not sure whether the latter would be a breaking change.

Adding schema validation would be a breaking change.

Conversation related to this could be found here -> https://coreos.slack.com/archives/G3T7N42NP/p1625153985037600?thread_ts=1625153575.035700&cid=G3T7N42NP

timflannagan: Yeah, that's true, adding extra schema validation around this field could be viewed as a breaking change. I think I still need to wrap my head around this usage in OLM, and would prefer an approach where we validate a CatalogSource as an input vs. serializing and parsing nested fields: https://github.com/operator-framework/api/blob/master/pkg/validation/internal/csv.go#L32. I'll poke around that slack conversation later :+1:.

anik120 (approved):
/lgtm

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -242,6 +248,28 @@ func (c *CatalogSource) Poll() bool {
...
+	var updateStrategyMap map[string]map[string]string
+	if err = json.Unmarshal(data, &updateStrategyMap); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{}
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Setting the default value of %v. Error:%v", DefaultRegistryPollDuration, err)
+		defaultTime, parseErr := time.ParseDuration(DefaultRegistryPollDuration)
```
kevinrizza, line 261: Should we be setting a default time? If this is not set correctly, should we just default to stripping instead?

anik120: I don't have a strong preference either way. To me either approach let's users know there's been some issue. Setting a default time does take care of the scenario where the CatalogSource was added by a ci and the admin was expecting to get updates anyway by default (instead of going days/months without looking at the error and expecting updates to come in, and then the issue getting escalated when realization sets in that updates were not coming in).

dinhxuanvu:
I would like to ask why we add a field into a spec to display parsing error? Spec is not where error and statuses should be reported. If we want to report conditions/statuses, then add them into the status instead. Honestly, I feel it is better in long term to add [Conditions](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go) to CatalogSource so we can properly project condition status to CatalogSource instead of having to add one-off field into CRD.

DhritiShikhar:
> I would like to ask why we add a field into a spec to display parsing error? Spec is not where errors and statuses should be reported. If we want to report conditions/errors, then add them into the status instead. Honestly, I feel it is better in long term to add [Conditions](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go) to CatalogSource so we can properly project condition status to CatalogSource instead of having to add one-off field into CRD.

Agreed. @anik120 Are you okay if I remove the `ParsingError` field?

pkg/operators/v1alpha1/catalogsource_types.go
exdx: I'm inclined with treating it as a terminal error when reconciling the catalog source and clearly stating on the status the error in parsing the interval. Other behaviors like setting a default, or removing the interval, could be surprising for users, versus the immediate feedback when the catalog spec is invalid.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -97,6 +101,8 @@ type RegistryPoll struct {
 	// The catalog operator polls to see if a new version of the catalog source is available.
 	// If available, the latest image is pulled and gRPC traffic is directed to the latest catalog source.
 	Interval *metav1.Duration `json:"interval,omitempty"`
+	// +optional
+	ParsingError string `json:"parsingError,omitempty"`
```
exdx, line 105: nit: this type of error should be populated on the status of the catalog, as its related to the reconciliation status of the catalog object on-cluster. The spec is for users to declare what the catalog should consist of and should generally not be modified by controllers.

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -103,6 +103,8 @@ spec:
                         interval:
                           description: Interval is used to determine the time interval between checks of the latest catalog source version. The catalog operator polls to see if a new version of the catalog source is available. If available, the latest image is pulled and gRPC traffic is directed to the latest catalog source.
                           type: string
+                        parsingError:
+                          type: string
```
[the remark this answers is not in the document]
njhale: If we decide to include the `parsingError` field, it should probably be in `status`.

anik120:
Closing this in favor of #165 that builds off of this. Thanks @DhritiShikhar for laying the ground work for this.

PR #143, opened 2021-07-28
cmd/operator-verify: Update the validating manifests command

timflannagan:
Update the operator-verify command package:

- Update the `... manifest validate` command and ensure only a single
  argument is being passed to the executable. This will fix a panic
  being produced when attempting to access the first element on the
  arguments array.
- Update an error message that uses a format specifier but no variable
  was being passed.
- Update the use/short/long description so it's clearer as to what the
  expected input should be.

Update the .gitignore:

- Ensure the root bin/ directory is being excluded, which typically houses binaries generated from local go build operations.

### DIFF
--- a/cmd/operator-verify/manifests/cmd.go
+++ b/cmd/operator-verify/manifests/cmd.go
@@ -11,13 +11,17 @@ import (
 
 func NewCmd() *cobra.Command {
 	rootCmd := &cobra.Command{
-		Use:   "manifests",
-		Short: "Validates all manifests in a directory",
-		Long: `'operator-verify manifests' validates a bundle in the supplied directory
-and prints errors and warnings corresponding to each manifest found to be
-invalid. Manifests are only validated if a validator for that manifest
-type/kind, ex. CustomResourceDefinition, is implemented in the Operator
-validation library.`,
+		Use:   "manifests <manifestDir>",
+		Short: "Validate all manifests in a directory",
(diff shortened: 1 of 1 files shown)


cmd/operator-verify/manifests/cmd.go
```diff
@@ -11,13 +11,17 @@ import (
...
-		Use:   "manifests",
-		Short: "Validates all manifests in a directory",
-		Long: `'operator-verify manifests' validates a bundle in the supplied directory
-and prints errors and warnings corresponding to each manifest found to be
-invalid. Manifests are only validated if a validator for that manifest
-type/kind, ex. CustomResourceDefinition, is implemented in the Operator
-validation library.`,
+		Use:   "manifests <manifestDir>",
+		Short: "Validate all manifests in a directory",
+		Args:  cobra.ExactArgs(1),
```
timflannagan, line 16: As noted in the commit/PR description, we're blindly accessing the `args[0]` array element within validating that element even exists:

```go
...
func manifestsFunc(cmd *cobra.Command, args []string) {
	bundle, err := manifests.GetBundleFromDir(args[0])
	if err != nil {
		log.Fatalf("Error generating bundle from directory: %s", err.Error())
	}
```

 We can ensure that a single argument has been provided at the cobra.Command level.

camilamacedo86 (approved):
/lgtm

PR #144, opened 2021-08-04
Validate SA names do not match csv deployment spec

kevinrizza:
OLM is unable to properly handle the case where a bundle author
explicitly adds a hardcoded reference to a service account with a name
that matches the name of the service account defined by the deployment
spec in the CSV. This commit adds a validation method to ensure that the
name of the service account in a bundle does not match
deployment.Spec.Template.Spec.ServiceAccountName in the csv's spec

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -8,6 +8,8 @@ import (
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/validation/internal/bundle.go
```diff
@@ -26,9 +28,35 @@ func validateBundles(objs ...interface{}) (results []errors.ManifestResult) {
...
+		saNamesFromCSV[saName] = struct{}{}
+	}
+
+	// find any hardcoded service account objects are in the bundle, then check if they match any sa definition in the csv
+	var errs []errors.Error
+	for _, obj := range bundle.Objects {
+		sa := v1.ServiceAccount{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &sa); err == nil {
+			if _, ok := saNamesFromCSV[sa.Name]; ok {
+				errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("invalid service account found in bundle. sa name cannot match service account defined for deployment spec in CSV"), sa.Name))
```
timflannagan, line 52: This looks like it's missing the format specifier?

kevinrizza: yeah, you're right, it doesn't need the format call anymore

timflannagan: I think we could've also made this error message a package constant but I'm not super concerned about that.

PR #147, opened 2021-08-17
add maxOpenShiftVersion test for parsing full version string 

ankitathomas:
Signed-off-by: Ankita Thomas <ankithom@redhat.com>

Test to specify behavior when reading full version string with 0 patch and pre-release version


dinhxuanvu (approved):
/lgtm

PR #148, opened 2021-08-24
(fix) : Fix check to verify label versions when it is =v4.x

camilamacedo86:
**Description**
CVP reported the error in the check when the label is = v4.8. 
This PR fixes the error and add a test to cover this scenario.


pkg/validation/internal/community.go
```diff
@@ -328,3 +324,14 @@ func validateImageFile(checks CommunityOperatorChecks, deprecatedAPImsg string)
 	}
 	return checks
 }
+
+// cleanStringToGetTheVersionToParse will remove the expected characters for
+// we are able to parse the version informed.
+func cleanStringToGetTheVersionToParse(value string) string {
+	doubleCote := "\""
```
njhale, line 331: `doubleQuote`?

njhale, line 331: typo: `doubleCote` -> doubleQuote?

pkg/validation/internal/testdata/dockerfile_v4_8/bundle_without_label.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM scratch
```
njhale, line 1: nit: it would be nice if we could template these and stamp generate our fixtures at runtime instead of keeping a bunch of near-duplicates around

njhale (commented):
Aside: The fact that we have platform-specific validations in this repo always catches me by surprise. We should start thinking about how to factor them out.

/approve

pkg/validation/internal/community.go
```diff
@@ -328,3 +324,14 @@ func validateImageFile(checks CommunityOperatorChecks, deprecatedAPImsg string)
...
+	singleCote := "'"
+	value = strings.ReplaceAll(value, singleCote, "")
+	value = strings.ReplaceAll(value, doubleCote, "")
+	value = strings.ReplaceAll(value, "v", "")
+	return value
```
joelanford, line 336:
```suggestion
	value, err := strconv.Unquote(value)
	if err != nil {
		return "", err
	}
	return strings.TrimPrefix(value, "v"), nil
```

A couple of notes:
- `strconv.Unquote()` is a safe unquote that tells you via an error if the string was not correctly quoted
- It seems like `strings.TrimPrefix(value, "v")` is safer than just replacing all "v"s in a string. Otherwise something like this would be valid `"vvv4vv.vvv8vvv"`.


If it helps, I wrote this function last week to convert from the label format to a semver range and then do a comparison: https://github.com/joelanford/community-operators-prod/blob/d10cb2dd9afda91a0626c10be93d1b819587d4f2/builder/main.go#L341-L392

pkg/validation/internal/testdata/dockerfile_v4_8/bundle_without_label.Dockerfile
camilamacedo86: That would be nicer by sure. We need to clean up this implementation.

pkg/validation/internal/community.go
camilamacedo86: Hi @joelanford,

Not proud of this code at all. We need refractory it and we could centralize something.
Your suggestion is making the code fail and with invalid syntax. So, can we do that in future follow up?

pkg/validation/internal/community.go
```diff
@@ -328,3 +324,14 @@ func validateImageFile(checks CommunityOperatorChecks, deprecatedAPImsg string)
 	}
 	return checks
 }
+
+// cleanStringToGetTheVersionToParse will remove the expected characters for
+// we are able to parse the version informed.
+func cleanStringToGetTheVersionToParse(value string) string {
+	doubleQuote := "\""
+	singleCote := "'"
```
dinhxuanvu, line 332: I assume this is `singleQuote`.

dinhxuanvu (changes requested):

pkg/validation/internal/community.go
joelanford: Is my example failing just because it introduces a need to have the function signature return `(string, error)`?

Either way, I think we should find a way to use `strconv.Unquote` and `strings.TrimPrefix(value, "v")` so that we don't accidentally allow values we would consider to be invalid. 

Any time we are permissive with a value, we must _always_ be permissive from that point forward, which is a bummer. I encountered lots of cases like this implementing file-based configs (e.g. icons that don't match their mediatypes or that don't even have valid base64 data).

camilamacedo86: > Is my example failing just because it introduces a need to have the function signature return (string, error)?
No, off course not. Note that `strconv.Unquote` is returning `syntax error`. 

Note as well that we can have more than only v4.8. It can be for example v4.8-v4.9. 

We need cleanups here and other improvements by sure. I am working now on follow-ups.

PR #120, opened 2021-05-23
additionals and improvements to dealing with deprecated API(s)

camilamacedo86:
**Description**

- New CommunityValiator to check the bundles against the required criteria to dealing with deprecated APIs for CRDs and the bundles which will be published on the community-operators (https://github.com/operator-framework/community-operators/tree/master/community-operators).
- improvements on the operatorhub.io validator to check the possible deprecated APIs kinds mapped on the project

**Check the new optional validator called via SDK**

Screen Shot 2021-05-23 at 19 39 28

**Also, check its CLI helpers**

Screen Shot 2021-05-23 at 19 41 15

Screen Shot 2021-05-23 at 19 40 44

**Related PRs**
- SDK CLI and docs: https://github.com/operator-framework/operator-sdk/pull/4939
- OLM docs: https://github.com/operator-framework/olm-docs/pull/155
- Community Operator docs: WIP

PR #149, opened 2021-08-26
 (feat): community validator: cleanups, changes and improvements

camilamacedo86:
**Description**

- Make clear the desire to move it out of the operator framework/api in the long term
- Change the key option to allow receive the annotation or the dockerfile as requested
- Add option to allow pass only the range as well instead of the file
- Improve the test coverage
- check if max ocp version is compatible with OCP label range 

**Motivation**
- Some consumers would like to check it with the annotations, others with the index image and others only by informing a specific range
- Clarifies its stage in Alpha and addresses requests/feedbacks made so far

Note that still requires to follow up to allow us to pass an OCP/k8s version to be checked as well.

### DIFF
--- a/pkg/validation/internal/community.go
+++ b/pkg/validation/internal/community.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"encoding/json"
+	golangerrors "errors"
 	"fmt"
 	"github.com/blang/semver"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


pkg/validation/internal/community.go
```diff
@@ -13,30 +14,48 @@ import (
...
+
+// deprecateOcpLabelMsg1_22 returns the specific ocp label message which is valid only for 1.22/OCP 4.9
+const deprecateOcpLabelMsg1_22 = "this bundle is using APIs which were deprecated and " +
+	"removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. " +
+	"Migrate the APIs " +
+	"for %s or provide compatible version(s) via the labels. (e.g. LABEL %s='4.6-4.8')"
 
-// CommunityOperatorValidator validates the bundle manifests against the required criteria to publish
-// the projects on the community operators
+// AlphaOCPCommunityOperatorValidator validates the bundle manifests against the required criteria to publish
```
kevinrizza, line 35: Why are we specifically referencing openshift in this repository?

camilamacedo86: We discussed this subject in the past remember. We needed to create a validator to test the max OCP version which is only valid for OCP and it fails in the OCP community rules which seems to be a grey area.  

We have not been away so far to provide validators that could be used by any tool via a lib that address requirements more like vendor-specific. So, we agreed (https://github.com/operator-framework/api/pull/120) to move forward with. That seems a discussion where we have been trying to find a solution for like 2 yrs. 

However, one of the feedbacks were about the name and we make it clear that it is to check the criteria against  ocp community catalog. Also, it seems that we also need to make clear that it is alpha and our intention is to move it out. See that it is highlighted via this PR in the code and docs that will be generated for this validator. The other changes in this pr are based on the feedbacks as well.

a) Improve the logic to check the versions with the @joelanford suggestions
b) allow informing the annotations or the dockerfile when we want to check the OCP config label. 
c) allow to pass a range too and improve the coverage. 

Also, in the future, now that we have at least a need to address we could try to use it to solve this matter and decouple this test from here to another place. So, IHMO we have:

**Problem:** We do have not an API that can be used to address needs that are vendor-specific / OCP

**Suggested Solution:**

Decouple the operator-framework/api and creates a project that will respect its interfaces and only address checks which are vendor-specific  

Use Cases

- I am as a maintainer would like to be able to write custom checks and validators vendor-specific that respects the same interfaces used for the others in a specific project so that, we can easily ensure that operator-framework api has no vendor-specific logics as desirable
- I am as a maintainer would like to have the vendor-specific implementations also available to be used as a LIB in any project such as SDK, audit or OPM so that, I can easily keep the implementation maintained with an encapsulated logic and useful for any project.

camilamacedo86: Hi @kevinrizza I came back for the original name and keep alpha.

camilamacedo86:
It was moved to: https://github.com/redhat-openshift-ecosystem/ocp-olm-catalog-validator
Also, we have now: https://github.com/operator-framework/api/pull/196

PR #151, opened 2021-08-30
Checks for k8s 1.25 deprecated APIs

bentito:
Adds an externally accessible function to check for APIs to be removed in Kubernetes 1.25.

The list of removed APIs is here:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

### DIFF
--- a/pkg/validation/external/removed_apis.go
+++ b/pkg/validation/external/removed_apis.go
@@ -1,0 +1,43 @@
+package external
+
+import (
+	"github.com/operator-framework/api/pkg/manifests"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func GetRemovedAPIsOn1_25From(bundle *manifests.Bundle) map[string][]string {
+	deprecatedAPIs := make(map[string][]string)
+	for _, obj := range bundle.Objects {
+		switch usage := obj.GetObjectKind().(type) {
+		case *unstructured.Unstructured:
+			switch usage.GetAPIVersion() {
+			case "batch/v1beta1":
+				if usage.GetKind() == "CronJob" {
+					addDepUsage(deprecatedAPIs, usage)
+				}
+			case "discovery.k8s.io/v1beta1":
+				if usage.GetKind() == "EndpointSlice" {
+					addDepUsage(deprecatedAPIs, usage)
+				}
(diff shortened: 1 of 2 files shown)


pkg/validation/external/removed_apis.go
```diff
@@ -0,0 +1,43 @@
+package external
+
+import (
+	"github.com/operator-framework/api/pkg/manifests"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func GetRemovedAPIsOn1_25From(bundle *manifests.Bundle) map[string][]string {
```
camilamacedo86, line 8: Why not add this implementation with https://github.com/operator-framework/api/blob/master/pkg/validation/internal/removed_apis.go?

pkg/validation/external/removed_apis.go
```diff
@@ -0,0 +1,43 @@
...
+			}
+		}
+	}
+	return deprecatedAPIs
+}
+
+func addDepUsage(deprecatedAPIs map[string][]string, u *unstructured.Unstructured) map[string][]string {
+	deprecatedAPIs[u.GetKind()] = append(deprecatedAPIs[u.GetKind()], u.GetName())
+	return deprecatedAPIs
+}
```
camilamacedo86, line 43: Are you getting the name of the resource where the api was found? See that it is used in more places so, why not address the cleanup across the project?

pkg/validation/external/removed_apis.go
```diff
@@ -0,0 +1,43 @@
+package external
+
+import (
+	"github.com/operator-framework/api/pkg/manifests"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func GetRemovedAPIsOn1_25From(bundle *manifests.Bundle) map[string][]string {
+	deprecatedAPIs := make(map[string][]string)
+	for _, obj := range bundle.Objects {
```
camilamacedo86, line 10: This check has not been called in any validator so it will not be useful. 
See that we would need to ensure that it will be called via the operatorhub and the alpha-deprecated one. 

we also need to address improvements for these scenario checks such as make clear what are the k8s versions supported by the checks and some cleanups.

PR #152, opened 2021-09-02
Add validation for Kind and APIVersion

josefkarasek:
Signed-off-by: Josef Karasek <jkarasek@redhat.com>


pkg/validation/internal/csv.go
```diff
@@ -194,3 +196,15 @@ func validateInstallModes(csv *v1alpha1.ClusterServiceVersion) (errs []errors.Er
 	}
 	return errs
 }
+
+// validateVersionKind checks presense of GroupVersionKind.Verion and GroupVersionKind.Kind
```
estroz, line 200:
```suggestion
// validateVersionKind checks presence of GroupVersionKind.Version and GroupVersionKind.Kind
```

estroz (commented):
any reason to not ensure group is present?

josefkarasek:
> any reason to not ensure group is present?

`group` is rarely present

Sometimes we can find something like `apiVersion: operators.coreos.com/v1alpha1`

PR #156, opened 2021-09-16
feat: add new validator to only check the usage of deprecate apis

camilamacedo86:
**Description**
The goal of this PR is to allow users to check only the criteria regards the deprecation of APIs.
It only add a new specific validator and move the checks already implemented to be centralized on it.


pkg/validation/internal/removed_apis.go
```diff
@@ -2,14 +2,149 @@ package internal
...
+// the CSV to do the checks. So, the criteria is >=minKubeVersion. By last, if the minKubeVersion is not provided
+// then, we should consider the operator bundle is intend to work well in any Kubernetes version.
+// Then, it means that:
+//--optional-values="k8s-version=value" flag with a value => 1.16 <= 1.22 the validator will return result as warning.
+//--optional-values="k8s-version=value" flag with a value => 1.22 the validator will return result as error.
+//minKubeVersion >= 1.22 return the error result.
+//minKubeVersion empty returns a warning since it would mean the same of allow install in any supported version
+func validateDeprecatedAPIS(bundle *manifests.Bundle, versionProvided string) (errs, warns []error) {
+	// K8s version where the apis v1betav1 is no longer supported
+	const k8sVerV1betav1Unsupported = "1.22.0"
```
dinhxuanvu, line 88: Can we elevate these two const (`k8sVerV1betav1Unsupported` and `k8sVerV1betav1Deprecated`) to be outside of this function so they can be like `ocpVerV1beta1Unsupported`? And on that note, why do we have an OCP const on an upstream repo? Should we remove `ocpVerV1beta1Unsupported`?

dinhxuanvu: This is not a big deal on the k8s const comment but the OCP const is certainly out of place in this repo.

dinhxuanvu (commented):
Looking good. I just have a quick question on the comment.

dinhxuanvu (approved):
/lgtm

PR #160, opened 2021-10-02
operatorhub: add check for bundle name size

camilamacedo86:
**Description**
Add check for bundle name size.
It will fail afterwords in OLM and in runtime and will not provide a nice msg.
In this way, we can prevent these scenarios by checking when we lint the bundle.

### DIFF
--- a/pkg/validation/internal/operatorhub.go
+++ b/pkg/validation/internal/operatorhub.go
@@ -193,6 +193,12 @@ func checkBundleName(checks CSVChecks) CSVChecks {
 			"naming convention: <operator-name>.v<semver> e.g. memcached-operator.v0.0.1", checks.csv.Name))
 	}
 
+	if len(checks.csv.Name) > 63 {
+		checks.errs = append(checks.errs, fmt.Errorf("csv.metadata.Name %v has the size %d which is bigger than " +
+			"63 caracteres. The bundle name must be 63 characters or less because it will be used as k8s ownerref label " +
+			"which only allows max of 63 characters", checks.csv.Name, len(checks.csv.Name)))
+	}
+
 	return checks
 }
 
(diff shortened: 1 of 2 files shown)


camilamacedo86:
c/c @dinhxuanvu @exdx

pkg/validation/internal/operatorhub.go
```diff
@@ -193,6 +193,12 @@ func checkBundleName(checks CSVChecks) CSVChecks {
 			"naming convention: <operator-name>.v<semver> e.g. memcached-operator.v0.0.1", checks.csv.Name))
 	}
 
+	if len(checks.csv.Name) > 63 {
+		checks.errs = append(checks.errs, fmt.Errorf("csv.metadata.Name %v has the size %d which is bigger than " +
+			"63 caracteres. The bundle name must be 63 characters or less because it will be used as k8s ownerref label " +
```
exdx, line 198: nit: typo in the first `characters` as `caracteres`

pkg/validation/internal/operatorhub_test.go
```diff
@@ -255,6 +255,12 @@ func TestCheckBundleName(t *testing.T) {
 			warnStrings: []string{"csv.metadata.Name memcached-operator.v1.3 is not following the " +
 				"versioning convention (MAJOR.MINOR.PATCH e.g 0.0.1): https://semver.org/"},
 		},
+		{
+			name:        "should return a error when the bundle name is > 63",
```
exdx, line 259: The operatorhub io validator is an optional validator if I'm not mistaken. Do we want to enforce this name length rule with the CSV validator instead? If any user will end up with a broken on-cluster experience if the name of their CSV is too long maybe this should go under the default CSV or bundle validator.

camilamacedo86:
I am closing this one because the CSV validator is already checking it. E.g 

$ operator-sdk bundle validate bundle
ERRO[0000] Error: Value : (memcached-operatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperator.v0.0.1) metadata.name "memcached-operatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperatoroperator.v0.0.1" is invalid: must be no more than 63 characters

PR #163, opened 2021-10-07
Add bundle size validation

bentito:
Initial take on adding a validation of the bundle size. Please let me know how it is wrong ;-)


pkg/validation/internal/operatorhub.go
```diff
@@ -95,6 +97,14 @@ func validateBundleOperatorHub(bundle *manifests.Bundle, k8sVersion string) erro
 		return result
 	}
```
camilamacedo86, line 98: Would that be possible?
- We have an attr https://github.com/operator-framework/api/blob/master/pkg/manifests/bundleloader.go#L20 for the size?
- when we load we set the size 
- Then we check that in the BundleValidator https://github.com/operator-framework/api/blob/master/pkg/validation/internal/bundle.go#L28

Also, note that we need to cover the implementations with unit tests as it is done in _test.go files to ensure the checks.

PR #165, opened 2021-10-20
fix: add custom Unmarshaler for the CatalogSource UpdateStrategy 

exdx:
This PR catches invalid polling duration intervals on the catalog spec during unmarshaling and updates them to a default (15m) in that case. There are no additional errors or changes to the spec/status generated, since this change is entirely within the unmarshaling step only. 

Builds off #142 

Signed-off-by: Daniel Sover <dsover@redhat.com>

### DIFF
--- a/pkg/operators/v1alpha1/catalogsource_types.go
+++ b/pkg/operators/v1alpha1/catalogsource_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -10,9 +11,12 @@ import (
 )
 
 const (
-	CatalogSourceCRDAPIVersion = GroupName + "/" + GroupVersion
-	CatalogSourceKind          = "CatalogSource"
+	CatalogSourceCRDAPIVersion  = GroupName + "/" + GroupVersion
+	CatalogSourceKind           = "CatalogSource"
+	DefaultRegistryPollDuration = "15m"
 )
+
+var defaultPollingDuration, _ = time.ParseDuration(DefaultRegistryPollDuration)
 
 // SourceType indicates the type of backing store for a CatalogSource
 type SourceType string
@@ -90,6 +94,34 @@ type CatalogSourceSpec struct {
 // Currently only registry polling strategy is implemented
 type UpdateStrategy struct {
 	*RegistryPoll `json:"registryPoll,omitempty"`
+}
+
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface and thus supports custom unmarshaling for an updateStrategy.
+// UnmarshalJSON checks to see if the provided polling interval is valid. If so, use it, if not set it to a default interval.
+// This custom unmarshaling is required in the case where a user provides an invalid string for the polling interval.
+// An invalid interval causes informer operations to fail at the cache level since the unmarshalerDecoder cannot populate the concrete type.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) error {
+	var updateStrategy map[string]map[string]string
+	if err := json.Unmarshal(data, &updateStrategy); err != nil {
+		return err
+	}
+	if updateStrategy == nil {
+		return nil
+	}
+
+	registryPoll := &RegistryPoll{}
+	if rp, ok := updateStrategy["registryPoll"]; ok {
+		if i, ok := rp["interval"]; ok {
+			duration, err := time.ParseDuration(i)
+			if err != nil {
+				registryPoll.Interval = &metav1.Duration{Duration: defaultPollingDuration}
+			} else {
+				registryPoll.Interval = &metav1.Duration{Duration: duration}
+			}
+		}
+	}
+	u.RegistryPoll = registryPoll
+	return nil
 }
 
 type RegistryPoll struct {
--- a/pkg/operators/v1alpha1/catalogsource_test.go
+++ b/pkg/operators/v1alpha1/catalogsource_test.go
@@ -1,0 +1,66 @@
+package v1alpha1
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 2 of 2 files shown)


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -92,6 +96,30 @@ type UpdateStrategy struct {
...
+	if err := json.Unmarshal(data, &updateStrategy); err != nil {
+		return err
+	}
+	if updateStrategy == nil {
+		return nil
+	}
+
+	registryPoll := &RegistryPoll{}
+	if rp, ok := updateStrategy["registryPoll"]; ok {
+		duration, err := time.ParseDuration(rp["interval"])
```
timflannagan, line 112: Does this also require a nil check attempting to access a nested map key's value?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -92,6 +96,30 @@ type UpdateStrategy struct {
 	*RegistryPoll `json:"registryPoll,omitempty"`
 }
 
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+// UnmarshalJSON checks to see the provided polling interval is valid. If so, use it, if not set it to the default interval.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) error {
```
timflannagan, line 101: I think I can wrap my head around why this implementation is favorable, but it wasn't immediately clear to me - any value in beefing up the description, and linking to an issue/BZ for why this workaround is necessary?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -92,6 +96,30 @@ type UpdateStrategy struct {
...
+	var updateStrategy map[string]map[string]string
+	if err := json.Unmarshal(data, &updateStrategy); err != nil {
+		return err
+	}
+	if updateStrategy == nil {
+		return nil
+	}
```
timflannagan, line 108: Any idea on when we'd encounter the updateStrategy map being nil after the unmarshaling process?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -92,6 +96,30 @@ type UpdateStrategy struct {
...
+
+	registryPoll := &RegistryPoll{}
+	if rp, ok := updateStrategy["registryPoll"]; ok {
+		duration, err := time.ParseDuration(rp["interval"])
+		if err != nil {
+			registryPoll.Interval = &metav1.Duration{Duration: defaultPollingDuration}
```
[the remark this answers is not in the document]
exdx: I think passing that error along to the status is a good option, but I'm not sure how we can do that inside of this unmarshaling function. We do not have access to the status inside the custom unmarshaler. Is there an easy way to set the status in this context? I feel like the pattern is controllers should set the status, this implementation is almost like a mutating webhook that changes a parameter to a default.

It may surprise users but this is an edge-case to begin with. I think we can document it as part of the existing polling docs.

pkg/operators/v1alpha1/catalogsource_types.go
exdx: I think in this case it would return the zero value which is not actually what we want. Good catch.

pkg/operators/v1alpha1/catalogsource_types.go
exdx: My reasoning was the updateStrategy field is optional on the spec and is a pointer type, so if it's not listed on the spec unmarshaling would return an empty map

exdx: Actually, I think the unmarshaling would not occur at all, since the JSON key is missing from the spec if updateStrategy is not set, so this check might not be very useful. Still prevents indexing into a nil map though.

pkg/operators/v1alpha1/catalogsource_types.go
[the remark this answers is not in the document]
exdx: The other option is to simply strip out the invalid interval and set updateStrategy to nil, which is also possible. If that seems less surprising we could do that instead.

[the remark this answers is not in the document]
timflannagan: Yeah, that's a point - I'm not entirely sure how we'd bubble this back up except returning a custom error that we check at the call site (or whether that's even possible). I'm not sure I have a super strong opinion on either approach: documentation sounds reasonable, but you'd still run into the UX gap when debugging further, vs. let's try and follow the last surprising approach (i.e. https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Might be worth trying to get some other opinions?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -92,6 +96,34 @@ type UpdateStrategy struct {
...
+	if updateStrategy == nil {
+		return nil
+	}
+
+	registryPoll := &RegistryPoll{}
+	if rp, ok := updateStrategy["registryPoll"]; ok {
+		if i, ok := rp["interval"]; ok {
+			duration, err := time.ParseDuration(i)
+			if err != nil {
+				registryPoll.Interval = &metav1.Duration{Duration: defaultPollingDuration}
```
anik120, line 117: Why don't we just return a custom error here instead? The CatalogSource reconciler can then display the unmarshalling error on it's status condition.

timflannagan: I think I'm also a fan of just bubbling up a custom error at this point (also suggested the same thing in https://github.com/operator-framework/api/pull/165#discussion_r733916173). We could also use a custom error, bubble that up to the call site, and at the call site set the default registry interval if we're still interested in making the explicit decision to default to an interval during parsing failures. I think I'm largely in the corner we should avoid defaulting to an interval during a parsing failure, and instead, bubbling that up to the CatalogSource's status field.

@exdx @njhale Any thoughts?

anik120:
closing in favor of #169

PR #168, opened 2021-11-02
Improve ClusterServiceVersion schema

### LINKED ISSUE #167: Improve API documentation for CSV and PackageManifest
The API docs for CSV and PackageManifests are incomplete for the following properties:

CSV properties:
```
.spec.customresourcedefinitions.owned[].resources[]
.spec.customresourcedefinitions.required[].resources[]
.spec.apiservicedefinitions.owned[].resources[]
.spec.apiservicedefinitions.required[].resources[]
```

These all share the same description:
> APIResourceReference is a Kubernetes resource type used by a custom resource

I've searched through OpenShift, Kubernetes, and Operator Framework repos and documentation and cannot find any further documentation of an `APIResourceReference` schema.  The schema consists of three string-type properties: `kind`,  ` name`, and `version`. These properties have no description defined in the CSV schema.
The `resource[]` properties above are currently only consumed by the web console in order to list related resources on operand details pages, but we recently had a bug opened where this API was being misused, and adding some

TheRealJon:
Fixes #167
Add descriptions for kind, name, and version properties of APIResourceReference type as well as DisplayName, Description, Keywords, Maintainers, Provider, Links, and Icon properties of ClusterServiceVersionSpec.


TheRealJon:
I realize now after opening this that the manifests are generated files. First time contributing to an API repo like this, so I'll research a bit more and update this.

camilamacedo86 (commented):
/lgtm

TheRealJon:
@camilamacedo86 I rebased this and added a few more descriptions to the CSV API so it will need another review/lgtm

dinhxuanvu (approved):
/approve

camilamacedo86:
It seems blocked by the resolution in : https://github.com/operator-framework/operator-lifecycle-manager/issues/2767
The concern is it increase the size of the CRDs and API only allows 1MB.

exdx (commented):
This looks like a helpful PR, and does improve the descriptions on the CSV API, but there is an issue with the size of the CRD when applied via `kubectl apply` as the `last-applied-configuration` annotation becomes too large for the server to accept. Increasing the size of the openapi schema will further complicate this issue. 

The long term solution is to limit the depth of the codegen generated for the openapi spec, to prevent this issue from occurring. That being said I don't know whether we should prevent this PR from going in based on the issue, until the longer term fix is prepared.

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -279,12 +283,33 @@ type ClusterServiceVersionSpec struct {
...
+
+	// A list of organizational entities maintaining the operator.
+	// +optional
 	Maintainers               []Maintainer              `json:"maintainers,omitempty"`
+
+	// The publishing entity behind the operator.
+	// +optional
 	Provider                  AppLink                   `json:"provider,omitempty"`
+
+	// A list of links related to the operator
```
porridge, line 307:
```suggestion
	// A list of links related to the operator.
```

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -148,8 +148,12 @@ type APIServiceDescription struct {
 // APIResourceReference is a Kubernetes resource type used by a custom resource
 // +k8s:openapi-gen=true
 type APIResourceReference struct {
+	// Name of the CustomResourceDefinition for the referent, or an empty string when the referent
+	// is not a custom resource
 	Name    string `json:"name"`
+	// Kind of the referent
 	Kind    string `json:"kind"`
+	// API version of the referent
```
porridge, line 156: Full stops are missing in these comments too but at least it's consistent with the comment for the containing type...

camilamacedo86:
@msugakov, 

Your question/issue seems out of scope here, see;

>What is the point/benefit of providing k8s kinds in spec.customresourcedefinitions.owned.[*].resources at all? 
You need to inform what are the CRDs which are owned by our project so that OLM will be able to handle and install them. 
 
> In order to fix https://bugzilla.redhat.com/show_bug.cgi?id=2082491 , we need to specify that our operator manages "just some deployments"

That is not specified on this spec. If your Operator manages a deployment why are you trying to do that in the CSV? it probably should be done via the Operator implementation itself and its controllers.

TheRealJon:
@msugakov The `spec.customresourcedefinitions.owned.[*].resources` is purely a frontend API used by the console. We use this to list resources that are managed by the operator in the topology view.

TheRealJon:
@camilamacedo86 @exdx @porridge I've rebased this PR. Could we revisit getting this merged?

perdasilva:
@TheRealJon you might need to do a make manifests or make verify to fix the verify stage

tlwu2013:
looks good to me too, thanks for the PR @TheRealJon!

tlwu2013:
+1, lgtm too, thanks for your PR @TheRealJon !

TheRealJon:
@dinhxuanvu I brought this up in the k89s slack OLM dev channel to loop @grokspawn in. If both of you are in agreement, I think this is ready to merge.

pkg/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -148,8 +148,11 @@ type APIServiceDescription struct {
 // APIResourceReference is a Kubernetes resource type used by a custom resource
 // +k8s:openapi-gen=true
 type APIResourceReference struct {
-	Name    string `json:"name"`
-	Kind    string `json:"kind"`
+	// Name of the CustomResourceDefinition for the referent, or an empty string when the referent is not a custom resource.
```
joelanford, line 151: Is this `metadata.name` of the CRD? If possible, it might be nice to be a little more explicit about where these are supposed to come from if its a simple mapping from CRD fields.

TheRealJon: Yeah, this field is the confusing one. I can update to 'metadata.name' instead of 'Name'

grokspawn:
/hold cancel

releasing this since it looks like we had some good dialog and reached a consensus.

joelanford:
/lgtm

Thanks!

PR #169, opened 2021-11-03
(catsrc) add custom unmarshaller for registry poll interval

anik120:
This PR implements a custom unmarshaller for the CatalogSourceSpec.UpdateStrategy.
When a value that cannot be unmarshalled to *metav1.Duration was being passed to
UpdateStrategy.RegistryPoll.Interval, the catalogSource sync loops in the catalog
operator was stuck in an infinite loop instead of backing off and performing other
syncs. The custom unmarshaller tries to unmarshall the input to the field to
*metav1.Duration, and if it fails, it sets a default value of 15m for the field.
Otherwise it accepts the value passed to the field.

### DIFF
--- a/pkg/operators/v1alpha1/catalogsource_types.go
+++ b/pkg/operators/v1alpha1/catalogsource_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -10,8 +11,9 @@ import (
 )
 
 const (
-	CatalogSourceCRDAPIVersion = GroupName + "/" + GroupVersion
-	CatalogSourceKind          = "CatalogSource"
+	CatalogSourceCRDAPIVersion  = GroupName + "/" + GroupVersion
+	CatalogSourceKind           = "CatalogSource"
+	DefaultRegistryPollDuration = 15 * time.Minute
 )
 
 // SourceType indicates the type of backing store for a CatalogSource
@@ -36,6 +38,8 @@ const (
 	CatalogSourceConfigMapError ConditionReason = "ConfigMapError"
 	// CatalogSourceRegistryServerError denotes when there is an issue querying the specified registry server.
 	CatalogSourceRegistryServerError ConditionReason = "RegistryServerError"
+	// CatalogSourceIntervalInvalidError denotes if the registry polling interval is invalid.
+	CatalogSourceIntervalInvalidError ConditionReason = "InvalidIntervalError"
 )
 
 type CatalogSourceSpec struct {
@@ -96,7 +100,32 @@ type RegistryPoll struct {
 	// Interval is used to determine the time interval between checks of the latest catalog source version.
 	// The catalog operator polls to see if a new version of the catalog source is available.
 	// If available, the latest image is pulled and gRPC traffic is directed to the latest catalog source.
-	Interval *metav1.Duration `json:"interval,omitempty"`
+	RawInterval  string           `json:"interval,omitempty"`
+	Interval     *metav1.Duration `json:"-"`
+	ParsingError string           `json:"-"`
+}
+
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
+	type alias struct {
+		*RegistryPoll `json:"registryPoll,omitempty"`
+	}
+	us := alias{}
+	if err = json.Unmarshal(data, &us); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{
+		RawInterval: us.RegistryPoll.RawInterval,
+	}
+	duration, err := time.ParseDuration(registryPoll.RawInterval)
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Using the default value of %s instead. Error: %s", DefaultRegistryPollDuration, err)
+		registryPoll.Interval = &metav1.Duration{Duration: DefaultRegistryPollDuration}
+	} else {
+		registryPoll.Interval = &metav1.Duration{Duration: duration}
+	}
+	u.RegistryPoll = registryPoll
+	return nil
 }
 
 type RegistryServiceStatus struct {
--- a/pkg/operators/v1alpha1/types_test.go
+++ b/pkg/operators/v1alpha1/types_test.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+	"fmt"
 	"sort"
 	"testing"
 	"time"
@@ -206,5 +208,74 @@ func TestCatalogSource_Poll(t *testing.T) {
 	}
 	for i, tt := range table {
 		require.Equal(t, tt.result, table[i].catsrc.Poll(), table[i].description)
+	}
+}
+
+func TestUpdateStrategyUnmarshal(t *testing.T) {
+	type TestStruct struct {
+		UpdateStrategy UpdateStrategy `json:"updateStrategy,omitempty"`
+	}
+	validDuration, err := time.ParseDuration("45m")
+	if err != nil {
+		panic(fmt.Errorf("error parsing duration: %s", err))
+	}
+	defaultDuration, err := time.ParseDuration("15m")
+	if err != nil {
+		panic(fmt.Errorf("error parsing duration: %s", err))
+	}
+	tests := []struct {
+		name string
+		in   []byte
+		out  TestStruct
+		err  error
+	}{
+		{
+			name: "valid",
+			in:   []byte(`{"UpdateStrategy": {"registryPoll":{"interval":"45m"}}}`),
+			out: TestStruct{
+				UpdateStrategy{
+					&RegistryPoll{
+						RawInterval:  "45m",
+						Interval:     &metav1.Duration{Duration: validDuration},
+						ParsingError: "",
+					},
+				},
+			},
+		},
+		{
+			name: "invalid",
+			in:   []byte(`{"UpdateStrategy": {"registryPoll":{"interval":"19mError Code"}}}`),
+			out: TestStruct{
+				UpdateStrategy{
+					&RegistryPoll{
+						RawInterval:  "19mError Code",
+						Interval:     &metav1.Duration{Duration: defaultDuration},
+						ParsingError: "error parsing spec.updateStrategy.registryPoll.interval. Using the default value of 15m0s instead. Error: time: unknown unit \"mError Code\" in duration \"19mError Code\"",
+					},
+				},
+			},
+		},
+		{
+			name: "empty",
+			in:   []byte(`{"UpdateStrategy": {"registryPoll":{"interval":""}}}`),
+			out: TestStruct{
+				UpdateStrategy{
+					&RegistryPoll{
+						Interval:     &metav1.Duration{Duration: defaultDuration},
+						ParsingError: "error parsing spec.updateStrategy.registryPoll.interval. Using the default value of 15m0s instead. Error: time: invalid duration \"\"",
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := TestStruct{}
+			err := json.Unmarshal(tt.in, &s)
+			require.Equal(t, tt.out.UpdateStrategy.RawInterval, s.UpdateStrategy.RawInterval)
+			require.Equal(t, tt.out.UpdateStrategy.Interval, s.UpdateStrategy.Interval)
+			require.Equal(t, tt.out.UpdateStrategy.ParsingError, s.UpdateStrategy.ParsingError)
+			require.Equal(t, tt.err, err)
+		})
 	}
 }


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
 	// Interval is used to determine the time interval between checks of the latest catalog source version.
 	// The catalog operator polls to see if a new version of the catalog source is available.
 	// If available, the latest image is pulled and gRPC traffic is directed to the latest catalog source.
-	Interval *metav1.Duration `json:"interval,omitempty"`
+	Interval     *metav1.Duration `json:"interval,omitempty"`
+	ParsingError string           `json:"-"`
```
timflannagan, line 104: Does this field need to be exported?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+}
+
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
+	var updateStrategyMap map[string]map[string]string
+	if err = json.Unmarshal(data, &updateStrategyMap); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{}
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
```
timflannagan, line 114: Do we need to nil check these fields to avoid blindly indexing and producing a panic?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+	ParsingError string           `json:"-"`
+}
+
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
+	var updateStrategyMap map[string]map[string]string
+	if err = json.Unmarshal(data, &updateStrategyMap); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{}
```
timflannagan, line 113: Just a quick thought (and this can be ignored): it looks like you can avoid an if/else statement structure by defaulting the `Interval` field to the `defaultTime` duration, and only override that value when the parsing the duration provided doesn't provide a parsing ignore.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Setting the default value of %v. Error:%v", DefaultRegistryPollDuration, err)
+		defaultTime, parseErr := time.ParseDuration(DefaultRegistryPollDuration)
+		if parseErr != nil {
+			return parseErr
```
timflannagan, line 119: I didn't see this anywhere in the PR description, but the idea is we're going to bubble up the parsing error and handle this at whatever call site is responsible for handling this, correct?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
+	var updateStrategyMap map[string]map[string]string
+	if err = json.Unmarshal(data, &updateStrategyMap); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{}
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Setting the default value of %v. Error:%v", DefaultRegistryPollDuration, err)
```
njhale, line 116: We might want to use the `%s` print directive for the error if we want to keep some go-specific metadata out of the status.

pkg/operators/v1alpha1/types_test.go
```diff
@@ -208,3 +209,65 @@ func TestCatalogSource_Poll(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := TestStruct{}
+			err := json.Unmarshal(tt.in, &s)
+			require.Equal(t, tt.out.UpdateStrategy.Interval, s.UpdateStrategy.Interval)
+			require.Equal(t, tt.err, err)
+		})
```
njhale, line 271: we're missing checks that ensure `ParsingError` is set as expected.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+		defaultTime, parseErr := time.ParseDuration(DefaultRegistryPollDuration)
+		if parseErr != nil {
+			return parseErr
+		}
```
njhale, line 120: Is it possible to parse the default a single time, say in `init()`, then deepcopy `defaultTime` for subsequent calls?

Also, failing to parse a default value is likely a non-recoverable programming error -- I'd probably `panic` in that case.

pkg/operators/v1alpha1/catalogsource_types.go
njhale, line 114: Do we know that these keys are always populated in the map access here? If not, we run the risk of a panic without checking for them first; i.e. `registryPollMap, ok := updateStrategyMap["registryPoll"]`

pkg/operators/v1alpha1/catalogsource_types.go
njhale: Yes, since we now want to use it outside of the `v1alpha1` package.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,30 @@ type RegistryPoll struct {
...
+func (u *UpdateStrategy) UnmarshalJSON(data []byte) (err error) {
+	var updateStrategyMap map[string]map[string]string
+	if err = json.Unmarshal(data, &updateStrategyMap); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{}
+	duration, err := time.ParseDuration(updateStrategyMap["registryPoll"]["interval"])
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Setting the default value of %v. Error:%v", DefaultRegistryPollDuration, err)
+		defaultTime, parseErr := time.ParseDuration(DefaultRegistryPollDuration)
```
njhale, line 117: I think we will immediately run into problems if we overwrite the given value with a default.

Let's walk through how:
1. User creates a CatalogSource with an invalid interval
2. Informers detect the change and attempt unmarshalling
3. The custom unmarshaller detects the invalid interval, sets the parsing error and sets the default interval
4. Custom unmarshalled CatalogSource enter our sync handlers
5. Sync handlers see the parsing error and set the invalid interval status condition
6. Sync handlers update CatalogSource with:
  - __valid__ default interval
  - invalid interval status condition
7. Informers detect the change and attempt unmarshalling
8. __The custom unmarshaller sees a valid interval, doesn't set parseError__
9. Custom unmarshalled CatalogSource enter our sync handlers
10. __Sync handlers see no parsing error and drop the invalid interval parsing condition__
11. If the user blinked, they would have no idea that:
  -  the CatalogSource they just created had an invalid status
  -  the CatalogSource has a different interval than what they intended

pkg/operators/v1alpha1/catalogsource_types.go
anik120: Specifically in the catsrc sync loops where we'll check for this error before projecting it onto the status condition.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: Sure, but you'd also have to do everything around the default time (parse the `15m` string to `time.Duration`, format the error etc) by default too, even if the error is nil. Having the if/else seems cleaner and avoids unnecessary operations every time.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: So this is a different parsing error, i.e parsing of the default `15m` string, which we know isn't going to return an error. The parsing error we're interested in bubbling up is [this one](https://github.com/operator-framework/api/pull/169/files#diff-b3658ff1a2084a857606b3e2991b9784faf58fa5dda1a8d93ae666b3b517995dR115). i.e we can get away with a simple `defaultTime, _ := time.ParseDuration(DefaultRegistryPollDuration)`, and that won't require the check `if parseErr != nil`, still it's a good idea to keep this check in in case the hardcoded `15m` is every tinkered around with with values that are not parseable to `time.Duration`

pkg/operators/v1alpha1/catalogsource_types.go
anik120: Sgtm

pkg/operators/v1alpha1/catalogsource_types.go
anik120: That is a great catch @njhale . I've updated the struct to include a `RawInterval` field that'll contain the user input, while the current `Interval` field will hold the parsed duration. Caveat is that if the default is set, the user will still see the ill format-ed input in the catsrc spec, but the status condition will contain a message stating that the default is being used instead. This looks like our best option in terms of what's least surprising for our users.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: @timflannagan @njhale I've changed this to unmarshall into an `alias` struct, so that we don't have to use a nested map and then blindly index that map for the value.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: `defaultTime` is only being parse once, here. There are no subsequent calls. 

I've changed the return statement to `panic` though.

pkg/operators/v1alpha1/catalogsource_types.go
anik120: Changed this statement to `panic` instead acc to @njhale 's suggestion.

pkg/operators/v1alpha1/catalogsource_types.go
perdasilva: Is there any reason to set a default at all? Couldn't we leave the catalog in an inoperable state until it's fixed?

pkg/operators/v1alpha1/types_test.go
```diff
@@ -208,3 +209,72 @@ func TestCatalogSource_Poll(t *testing.T) {
...
 	}
 }
+
+func TestUpdateStrategyUnmarshal(t *testing.T) {
+	type TestStruct struct {
+		UpdateStrategy UpdateStrategy `json:"updateStrategy,omitempty"`
+	}
+	validDuration, err := time.ParseDuration("45m")
+	if err != nil {
+		return
```
perdasilva, line 219: will the test succeed on this premature return?

anik120: @perdasilva good catch, a hold up from the original PR, I've changed it to `panic` instead.

pkg/operators/v1alpha1/catalogsource_types.go
timflannagan: @anik120 This is likely an artifact of too many context switches to apologies if you've already answered this: can we simply just bubble up the registryPoll.ParsingError without setting a default? I was playing around with this locally by manually updating the o-f/api vendored code in my OLM fork and was surprised by the behavior (although it did fix the problematic behavior of failing to establish a watch).

pkg/operators/v1alpha1/catalogsource_types.go
njhale: > defaultTime is only being parse once, here. There are no subsequent calls.

IIUC, it's called every time the custom marshaller is used, which is every time the resource is "read off the wire" (really constructed from the delta-fifo queue for most cases, but that's nitpicking)

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -10,8 +11,9 @@ import (
 )
 
 const (
-	CatalogSourceCRDAPIVersion = GroupName + "/" + GroupVersion
-	CatalogSourceKind          = "CatalogSource"
+	CatalogSourceCRDAPIVersion  = GroupName + "/" + GroupVersion
+	CatalogSourceKind           = "CatalogSource"
+	DefaultRegistryPollDuration = "15m"
```
njhale, line 16: Why isn't this just a `time.Duration`?

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -36,6 +38,8 @@ const (
 	CatalogSourceConfigMapError ConditionReason = "ConfigMapError"
 	// CatalogSourceRegistryServerError denotes when there is an issue querying the specified registry server.
 	CatalogSourceRegistryServerError ConditionReason = "RegistryServerError"
+	// CatalogSourceIntervalInvalidError denotes if the registry polling interval is invalid.
+	CatalogSourceIntervalInvalidError ConditionReason = "IntervalInvalidError"
```
njhale, line 42: nit: `InvalidIntervalError` sounds prettier

pkg/operators/v1alpha1/catalogsource_types.go
njhale: > Is there any reason to set a default at all? Couldn't we leave the catalog in an inoperable state until it's fixed?

My memory is a little hazy around this too. On the surface, I'm with @perdasilva and @timflannagan here, unless @anik120 could provide some reasoning for the default in the invalid case.

anik120: @perdasilva @njhale @timflannagan so getting an error in unmarshalling the value is what started this fix in the first place. If there is no custom unmarshaller, a value like "15mError.code" gets the controller stuck in a loop trying to set up a watch for the catsrc. If we return an error from the custom unmarshaller, it takes us back to the same problem, the controller gets stuck in a loop trying to set up a watch for the catsrc. Since the correct way to avoid this is schema validation, but that is not back portable in this case, handling this error in this way and setting a default is the only correct solution for this problem.

anik120: Also, since we get into the same problem of failing to set up a watch for the catsrc, none of the syncs are performed for the catsrc, so the catalog operator logic of checking for the `parsingError` and setting the status condition if there is an error is never executed. i.e the users are left in the dark about why the catsrc they added did not do anything for them, which forces them to inspect the pod logs, which is another thing we are trying to avoid.

anik120: Here's the PR: https://github.com/operator-framework/operator-lifecycle-manager/pull/2447

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,32 @@ type RegistryPoll struct {
 	// Interval is used to determine the time interval between checks of the latest catalog source version.
 	// The catalog operator polls to see if a new version of the catalog source is available.
 	// If available, the latest image is pulled and gRPC traffic is directed to the latest catalog source.
-	Interval *metav1.Duration `json:"interval,omitempty"`
+	RawInterval  string           `json:"interval,omitempty"`
+	Interval     *metav1.Duration `json:"-"`
+	ParsingError string           `json:"-"`
```
awgreene, line 105: Why is the Parsing error written to the spec? I assume its because it isn't possible to bubble this up and write it to the status?

I understand that following suggestion may not be viable for a bug fix, but I strongly believe that this would be better addressed through use of an webhook that sets sane defaults or rejects invalid requests, as described [here](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html).

awgreene (changes requested):
I do not believe that this is the correct way to address the bug. However, given the ongoing plan to introduce new APIs, I could be swayed to slap an approval label on this.

pkg/operators/v1alpha1/catalogsource_types.go
njhale: If a CatalogSource already exists with an invalid interval, what happens to it when the new version of OLM, containing said webhook, is installed?

The `json:"-"` tag indicates the field shouldn't be (un)marshaled, so they're not actually visible to clients of the k8s API. The end effect is that the end user/client both sees no new fields and loses no old fields; on the flip-side, OLM gets a new field to persist extra internal info about field parsing issues in memory.

This is very backportable.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -96,7 +100,32 @@ type RegistryPoll struct {
...
+	us := alias{}
+	if err = json.Unmarshal(data, &us); err != nil {
+		return err
+	}
+	registryPoll := &RegistryPoll{
+		RawInterval: us.RegistryPoll.RawInterval,
+	}
+	duration, err := time.ParseDuration(registryPoll.RawInterval)
+	if err != nil {
+		registryPoll.ParsingError = fmt.Sprintf("error parsing spec.updateStrategy.registryPoll.interval. Using the default value of %s instead. Error: %s", DefaultRegistryPollDuration, err)
```
tylerslaton, line 122: nit: Should we build this into an error here and return it?

pkg/operators/v1alpha1/types_test.go
```diff
@@ -208,3 +210,72 @@ func TestCatalogSource_Poll(t *testing.T) {
...
+					&RegistryPoll{
+						RawInterval:  "19mError Code",
+						Interval:     &metav1.Duration{Duration: defaultDuration},
+						ParsingError: "error parsing spec.updateStrategy.registryPoll.interval. Using the default value of 15m0s instead. Error: time: unknown unit \"mError Code\" in duration \"19mError Code\"",
+					},
+				},
+			},
+		},
+		{
+			name: "empty",
```
tylerslaton, line 259: nit: should we add a test case if an empty JSON blob is sent?

tylerslaton (approved):
A couple of small nits. I'll LGTM in a bit just to make sure we have time to look at them.

pkg/operators/v1alpha1/types_test.go
anik120: Is an empty JSON blob for an `Unmarshaller` func ever an input? The unmarshaller func only gets called when there is valid data to unmarshall (i.e if you don't specify `UpdateStrategy` in your CatalogSource, this function is not getting invoked by the complier anyway)

tylerslaton: I see, thank you for the clarification.

tylerslaton:
Nice, great work!

/lgtm

PR #171, opened 2021-11-08
doc: fix layout of OperatorHub doc

camilamacedo86:
**Description**
Just fix the layout for the doc: https://pkg.go.dev/github.com/operator-framework/api@master/pkg/validation/internal#OperatorHubValidator

### DIFF
--- a/pkg/validation/internal/operatorhub.go
+++ b/pkg/validation/internal/operatorhub.go
@@ -22,33 +22,55 @@ import (
 // the projects on OperatorHub.io.
 //
 // This validator will ensure that:
-// - The annotations capabilities into the CSV has a valid option, which are:
-// * "Basic Install"
-// * "Seamless Upgrades"
-// * "Full Lifecycle"
-// * "Deep Insights"
-// * "Auto Pilot"
-// - The annotations categories into the CSV has a valid option, which are:
-// * "AI/Machine Learning"
-// * "Application Runtime"
-// * "Big Data"
-// * "Cloud Provider"
-// * "Developer Tools"
-// * "Database"
-// * "Integration & Delivery"
-// * "Logging & Tracing"
-// * "Modernization & Migration"
-// * "Monitoring"
-// * "Networking"
-// * "OpenShift Optional"
-// * "Security"
-// * "Storage"
-// * "Streaming & Messaging"
 //
-// NOTE: The operatorhub validator can verify against custom bundle categories by setting the OPERATOR_BUNDLE_CATEGORIES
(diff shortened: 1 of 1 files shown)


pkg/validation/internal/operatorhub.go
```diff
@@ -22,33 +22,55 @@ import (
...
+// * Basic Install
+//
+// * Seamless Upgrades
+//
+// * Full Lifecycle
+//
+// * Deep Insights
+//
+// * Auto Pilot
+//
```
tlwu2013, line 37: +1 on this added spacing since it helps separate "capabilities" above and the "categories" below.

tlwu2013 (commented):
I am not sure if it's just me but adding all these extra spacings makes me feel it's a bit harder to read through the comments.

I do like some extra spacings to better separate the zoning though. So perhaps the rule could be we add spacings between every `-` but no extra spacings between `*` so things are that not wide spread but still provides some zoning? 

What do you think?

camilamacedo86:
Hi @tlwu2013, 

> I am not sure if it's just me but adding all these extra spacings makes me feel it's a bit harder to read through the comments.

The changes here are required to fix the layout for its docs to be generated properly. We ought to not ask for a look in the code and instead provide the link: https://pkg.go.dev/github.com/operator-framework/api@master/pkg/validation/internal#OperatorHubValidator 

> I do like some extra spacings to better separate the zoning though. So perhaps the rule could be we add spacings between every - but no extra spacings between * so things are that not wide spread but still provides some zoning?

The spaces are to allow breaking the lines in the golang docs generated for the API. adopted - instead of * will not solve this need.

tlwu2013:
thanks for the clarification, sgtm @camilamacedo86

pkg/validation/internal/operatorhub.go
```diff
@@ -22,33 +22,55 @@ import (
 // the projects on OperatorHub.io.
 //
 // This validator will ensure that:
+//
 // - The annotations capabilities into the CSV has a valid option, which are:
-// * "Basic Install"
-// * "Seamless Upgrades"
-// * "Full Lifecycle"
-// * "Deep Insights"
-// * "Auto Pilot"
```
ankitathomas, line 30: indentation also works for correcting the output formatting
```
// - The annotations capabilities into the CSV has a valid option, which are:
//   * "Basic Install"
//   * "Seamless Upgrades"
//   * "Full Lifecycle"
//   * "Deep Insights"
//   * "Auto Pilot"
```

ankitathomas (approved):
/lgtm

PR #173, opened 2021-11-16
feat: update catalog source spec with config

perdasilva:
Signed-off-by: Per G. da Silva <perdasilva@redhat.com>

### Description
This PR adds a new optional `grpcPodConfig` field to the `CatalogSource` spec. The config includes: node selector and toleration configuration, which will eventually make its way to the CatalogSource pod. With this change, and subsequent catalog operator update, users will be able to specify node selectors and tolerations for the underlying catalog source pod.

### Testing

I need some guidance here if we have a test suite to exercise the creation of the objects. So far, I've manually tested the creation against the spec.


perdasilva:
Refrain from merging this until we've all aligned on https://github.com/operator-framework/operator-lifecycle-manager/issues/2452

perdasilva:
It seems after the discussion this may not be the right approach. Closing the PR for now.

perdasilva:
@kevinrizza @anik120 wdyt? I've tried to keep the number of exposed knobs to a minimum. Is there anything else you guys think we need in there?

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -67,6 +67,41 @@ spec:
                 displayName:
                   description: Metadata
                   type: string
+                grpcPodConfig:
+                  description: GrpcPodConfig exposes different overrides for the pod spec of the CatalogSource Pod. Only used when SourceType = SourceTypeGrpc and Image is set.
+                  type: object
+                  properties:
+                    nodeSelector:
+                      description: 'NodeSelector is a selector which must be true for the pod to fit on a node. Selector which must match a node''s labels for the pod to be scheduled on that node. More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/'
```
anik120, line 75: I can see in the comment above the struct field that you used the single quote, not sure why it rendered it here as double quotes. 
```suggestion
                      description: 'NodeSelector is a selector which must be true for the pod to fit on a node. Selector which must match a node's labels for the pod to be scheduled on that node. More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/'
```

anik120 (commented):
@perdasilva I think this list looks good. This should be more than enough for now, and we can always add more if we need later, although I haven't seen any request for anything more than what you've included so far.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -86,6 +92,28 @@ type CatalogSourceSpec struct {
...
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Tolerations are the catalog source's pod's tolerations.
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// If specified, indicates the pod's priority. "system-node-critical" and
```
kevinrizza, line 107: nit: This comment feels ripe for rot to me. Do we need to call out the PriorityClass implementation details as they exist today?

kevinrizza (approved):
This looks good to me, just a nit about code comments

crds/operators.coreos.com_catalogsources.yaml
perdasilva: I've removed the reference to external documentation. That fixed the underlying issue.

pkg/operators/v1alpha1/catalogsource_types.go
perdasilva: Great call out! I've updated!

perdasilva:
Thank you @kevinrizza and @anik120! I've addressed your comments ^^

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -69,6 +70,11 @@ type CatalogSourceSpec struct {
 	// +Optional
 	Image string `json:"image,omitempty"`
 
+	// GrpcPodConfig exposes different overrides for the pod spec of the CatalogSource Pod.
+	// Only used when SourceType = SourceTypeGrpc and Image is set.
+	// +Optional
```
joelanford, line 75: Is the `+Optional` flag case insensitive? I've only ever seen it written as `+optional`.

joelanford (approved):
/lgtm

pkg/operators/v1alpha1/catalogsource_types.go
perdasilva: That's a great question. I just kept what was around. I've updated everything to be lowercase to keep it with the general lower-case convention.

PR #174, opened 2021-11-16
fix: clarifies message error regards service account

camilamacedo86:
**Description**
Clarifies error message for service account

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -52,7 +52,10 @@ func validateServiceAccounts(bundle *manifests.Bundle) []errors.Error {
 		sa := v1.ServiceAccount{}
 		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &sa); err == nil {
 			if _, ok := saNamesFromCSV[sa.Name]; ok {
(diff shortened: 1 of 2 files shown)


pkg/validation/internal/bundle.go
```diff
@@ -52,7 +52,10 @@ func validateServiceAccounts(bundle *manifests.Bundle) []errors.Error {
 		sa := v1.ServiceAccount{}
 		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &sa); err == nil {
 			if _, ok := saNamesFromCSV[sa.Name]; ok {
-				errs = append(errs, errors.ErrInvalidBundle("invalid service account found in bundle. sa name cannot match service account defined for deployment spec in CSV", sa.Name))
+				errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("invalid service account found in bundle. " +
+					"OLM cannot handle more than one Service Account defined with the same name. " +
```
kevinrizza, line 56: This error message is probably going to also be confusing for users. It's not really that OLM can't handle multiple service accounts with more than one name, it's that kubernetes doesn't let you name multiple service accounts with the same name.

Something like

"This service account $SANAME in your bundle is not valid, because a service account with the same name was already specified in your CSV. If this was unintentional, please remove the service account manifest from your bundle. If it was intentional to specify a separate service account, please rename the SA in either the bundle manifest or the CSV."

PR #175, opened 2021-11-17
Add WarnState

theishshah:
Add WarnState for Scorecard result types

### DIFF
--- a/pkg/apis/scorecard/v1alpha3/test_types.go
+++ b/pkg/apis/scorecard/v1alpha3/test_types.go
@@ -14,6 +14,8 @@ const (
 	FailState State = "fail"
 	// ErrorState occurs when a Test encounters a fatal error and the reported points should not be considered.
 	ErrorState State = "error"
+	//WarnState occurs when a Test's ExpectedPoints == 0 && Test is not required for functionality
+	WarnState State = "warn"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/apis/scorecard/v1alpha3/test_types.go
```diff
@@ -14,6 +14,8 @@ const (
 	FailState State = "fail"
 	// ErrorState occurs when a Test encounters a fatal error and the reported points should not be considered.
 	ErrorState State = "error"
+	//WarnState occurs when a Test's ExpectedPoints == 0 && Test is not required for functionality
```
joelanford, line 17: What dictates whether a test is required or optional?

camilamacedo86: Is not the warnState defined by the suggestion : https://github.com/operator-framework/api/blob/d59f62ff6ac73717b98f2ba7be5bb854f80d8910/pkg/apis/scorecard/v1alpha3/test_types.go#L32 ? 

IF we add that what will be the purposed of the suggestion one?

joelanford:
/hold

Let's discuss if a Warn level is necessary vs just using the existing suggestions field, and changing these tests to never fail.

perdasilva:
closing as stale

Issue #172, opened 2021-11-09
Add common linter check to ensure that all operator/managers defined the resource request


camilamacedo86:
## What

Verify that all the operators' Deployment has resource requests set on every container. In cases where no resource requests have been supplied for CPU and memory the verification should fail, e.g;

```yml
  ...
  # TODO(user): Configure the resources accordingly based on the project requirements.
  # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
  resources:
    requests:
      cpu: 10m
      memory: 64Mi
    ... 
```

## Why

Kubernetes resource requests are required for ResourceQuota to work. Users would like to apply these quotas in namespaces where operators are deployed, to limit consumption on the cluster cores.

## Where

Create a new validator to check such as (ResourcesRequestsValidator). 

c/c @dmesser

camilamacedo86:
Hi, @tlwu2013 and @dmesser, 

So, concerns were raised against adding the validator to enforce/fail when the requested resource is not set. In this way, I'd like to add here and share the mainly strongest argumentations for both cases:  

The motivation to mandatory ask for authors to set at least the resource request is: 

- If a LimitRange is activated in a namespace for computing resources like CPU and memory, users must specify requests or limits for those values. Otherwise, the system may reject Pod creation.
- If a resourQuota be configured and the pods do not have at least requests or limits for those values the system may reject Pod creation.

And then, the above configuration is commonly adopted and recommended in multi-tenancy and or production clusters where OLM is used usually. More info: https://github.com/operator-framework/operator-sdk/pull/5353/files

The argumentation against making the request resource mandatory is: 
- `We want these operator authors to set GOOD resource requests. To create good resource requests, operator authors will need time and customer usage data`. However, the Operator Authors are able to do tests and benchmarks before publishing their first solution or indeed adopting reasonable batting started values for the requests such as: 

```
  resources:
    requests:
      cpu: 10m
      memory: 64Mi
```
- Usually, we only make mandatory what will explicitly not work well with OLM and that shows to be more an indirectly need/requirement. 

Additionally, an additional idea raise so far was to add giving operators a "badge" in the UI when the operator meets these criteria. 

However, it also shows that everybody agrees that the Operator Author should be hammered with warnings.

IHMO: The argumentation to add the requested resource as mandatory seems stronger. But, we can also get started by warning and be revisiting it in the future as well.  

c/c @kevinrizza @awgreene @fgiloux

dmesser:
[...]

PR #197, opened 2021-11-28
feat: add new validator to check/warning when the resource request is not defined for a container used/shipped in the bundle

camilamacedo86:
**Description**
Add a new validator that will check the resources of the containers shipped on the bundle

**Motivation**
Closes: https://github.com/operator-framework/api/issues/172


pkg/validation/internal/resources.go
```diff
@@ -0,0 +1,76 @@
...
+
+	for _, dSpec := range deploymentSpec {
+		for _, c := range dSpec.Spec.Template.Spec.Containers {
+			if c.Resources.Requests == nil || !(len(c.Resources.Requests.Cpu().String()) != 0 && len(c.Resources.Requests.Memory().String()) != 0) {
+				msg := fmt.Errorf("it is recommended set the resource requests for all containers " +
+					"created and managed via the Operator project. Note that the container %s does not " +
+					"have a resource request for CPU and/or Memory define. Inform the minimal general values required " +
+					"to ensure that the containers will have them on the cluster. " +
+					"More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/", c.Name)
+				warns = append(warns, msg)
```
camilamacedo86, line 71: /hold

Memory and CPU requests and limits are associated with Containers, but be aware that the Memory and CPU 
requests and limits of a Pod are the sums of its specific computing types for all the containers in the Pod. 

We need to change it to be the sum

camilamacedo86: /hold cancel 

@fgiloux dis some checks and:

> But if a LimitRange constraint at the Pod level or a Quota is specified all the containers of a pod need to provide resource requests or limits otherwise the Pod gets rejected

So, the validator here shows right we need to warn for all containers.

pkg/validation/internal/good_practices.go
```diff
@@ -0,0 +1,76 @@
...
+	goerrors "errors"
+	"fmt"
+
+	"github.com/operator-framework/api/pkg/manifests"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
+// GoodPracticesValidator validates the bundle against criteria and suggestions define as
```
perdasilva, line 13:
```suggestion
// GoodPracticesValidator validates the bundle against criteria and suggestions defined as
```

perdasilva: Could we add a documentation link?

pkg/validation/internal/good_practices.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// - The resources request for CPU and/or Memory are not defined for any of the containers found in the CSV
+var GoodPracticesValidator interfaces.Validator = interfaces.ValidatorFunc(goodPracticesValidator)
+
+func goodPracticesValidator(objs ...interface{}) (results []errors.ManifestResult) {
+	for _, obj := range objs {
+		switch v := obj.(type) {
+		case *manifests.Bundle:
+			results = append(results, validateGoodPracticesFrom(v))
+		}
```
perdasilva, line 26: Should we warn or panic on the default case?

pkg/validation/internal/good_practices_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	tests := []struct {
+		name        string
+		args        args
+		wantError   bool
+		wantWarning bool
+		errStrings  []string
+		warnStrings []string
+	}{
+		{
```
perdasilva, line 21: Is it th exercising the other error paths as well, e.g. when the bundle is nil, or the bundle.CSV is nil, etc.

perdasilva: also when `csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs` is nil

kevinrizza:
/approve

This looks reasonable to me. Per's comments seem relevant, but other than that I'm fine with merging this.

pkg/validation/internal/good_practices.go
camilamacedo86: That is true. We have not in any of the validators.
Tried to cover all scenarios but that seems out of the scope.

pkg/validation/internal/good_practices.go
camilamacedo86: added

pkg/validation/internal/good_practices.go
camilamacedo86: No, because that would mean panic or warning when we are checking all objects looking for the bundle. (see that it is a for)

To ensure that we always will have the bundle here it is checked before in the default checks made.

pkg/validation/internal/good_practices_test.go
camilamacedo86: Ideally yes, I agree.
Good point done. 🥇 
thank you

pkg/validation/internal/good_practices.go
perdasilva: ah ok - you're right! I get it now.

PR #198, opened 2021-11-29
Introduce OLMConfig CRD

awgreene:


pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Features contains the list of configurable OLM features.
+type Features struct {
+	// DisableCopiedCSVs is used to disable OLM's "Copied CSV" feature
+	// for operators installed at the cluster scope, where a cluster
+	// scoped operator is one that has been installed in an
+	// OperatorGroup that targets all namespaces.
+	// When reenabled, OLM will recreate the "Copied CSVs" for each
+	// cluster scoped operator.
```
awgreene, line 19: Should add a default value of false here.

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// OLMConfigList is a list of OLMConfig resources.
+type OLMConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	// +listType=set
+	Items []OperatorGroup `json:"items"`
```
awgreene, line 50:
```suggestion
	Items []OLMConfig `json:"items"`
```

pkg/operators/v1/olmconfig_types.go
awgreene: Done in the latest version of the PR.

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,58 @@
...
+type Features struct {
+	// +kubebuilder:default=false
+
+	// DisableCopiedCSVs is used to disable OLM's "Copied CSV" feature
+	// for operators installed at the cluster scope, where a cluster
+	// scoped operator is one that has been installed in an
+	// OperatorGroup that targets all namespaces.
+	// When reenabled, OLM will recreate the "Copied CSVs" for each
+	// cluster scoped operator.
+	DisableCopiedCSVs bool `json:"disableCopiedCSVs,omitempty"`
```
timflannagan, line 22: You may need to make this a pointer to a bool to avoid `disableCopiedCSVs: false` being implicitly specified (or printed when firing off `kubectl get olmconfigs <...> -o yaml`).

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// OLMConfigStatus is the status for an OLMConfig resource.
+type OLMConfigStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
timflannagan, line 27: Does this need to be `//+optional`?

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:storageversion
+// +kubebuilder:resource:categories=olm,scope=Cluster
```
timflannagan, line 34: Should this resource be cluster scoped?

awgreene: Based on the original [enhancement](https://github.com/operator-framework/enhancements/pull/92/files#diff-c21d583efb087d4f578e5a734ef0e50615eec9743cb5ffa9fb388f32041bc6b9R105) it should be.

pkg/operators/v1/olmconfig_types.go
awgreene: The CRD Generation tool infers that it is options by the `omitempty` tag.

pkg/operators/v1/olmconfig_types.go
awgreene: Ha! Actually landed on this change while consuming the API. Done!

crds/operators.coreos.com_olmconfigs.yaml
```diff
@@ -0,0 +1,97 @@
...
+                      type: boolean
+                      default: false
+            status:
+              description: OLMConfigStatus is the status for an OLMConfig resource.
+              type: object
+              properties:
+                conditions:
+                  type: array
+                  items:
+                    description: "Condition contains details for one aspect of the current state of this API Resource. --- This struct is intended for direct use as an array at the field path .status.conditions.  For example, type FooStatus struct{     // Represents the observations of a foo's current state.     // Known .status.conditions.type are: \"Available\", \"Progressing\", and \"Degraded\"     // +patchMergeKey=type     // +patchStrategy=merge     // +listType=map     // +listMapKey=type     Conditions []metav1.Condition `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\" protobuf:\"bytes,1,rep,name=conditions\"` \n     // other fields }"
```
timflannagan, line 54: Hmm interesting - this appears to be rendered weird to me, but poking around other generated CRDs using this tooling, it largely resembles unrelated projects as well so maybe this is just something we just have to deal with?

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func init() {
+	SchemeBuilder.Register(&OLMConfig{}, &OLMConfigList{})
+}
+
+func (config *OLMConfig) CopiedCSVsAreEnabled() bool {
+	if config == nil {
+		return true
+	}
```
timflannagan, line 63: I think we can make a reasonable assumption that this kind of null check would be handled further up in the chain or directly at the call site, but it doesn't hurt making this (potentially duplicate) check :+1:.

crds/operators.coreos.com_olmconfigs.yaml
awgreene: This documentation is derived from the [metav1.Condition](https://github.com/kubernetes/apimachinery/blob/5f072755815add51938b52f806c555ed8c86a1ce/pkg/apis/meta/v1/types.go#L1446) struct and I do not believe that we can override it with yaml manipulation.

pkg/operators/v1/olmconfig_types.go
awgreene: Agreed, hopefully this code is never hit but it seemed like a nice safety net.

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return !config.Spec.Features.DisableCopiedCSVs
+}
```
joelanford, line 66: Is there a specific ConditionType we're going to use to signal that the `DisableCopiedCSVs` feature has been acknowledged/processed?

If so, should we capture the relevant condition type(s) and reasons here?

pkg/operators/v1/olmconfig_types.go
joelanford: This was marked as resolved, but noticed that its not (or no longer) a pointer to a bool?

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,66 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OLMConfigSpec is the spec for an OLMConfig resource.
+type OLMConfigSpec struct {
+	Features Features `json:"features,omitempty"`
```
joelanford, line 9: Should we make this optional as well? And make it a pointer?

pkg/operators/v1/olmconfig_types.go
awgreene: Good idea!

pkg/operators/v1/olmconfig_types.go
awgreene: Sure.

pkg/operators/v1/olmconfig_types.go
awgreene: Addressed in the latest version of the PR!

pkg/operators/v1/olmconfig_types.go
awgreene: Done!

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,71 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	DisableCopiedCSVsConditionType = "DisableCopiedCSVs"
```
joelanford: Nit: Should this be past tense so that it reads a little more like some state has (or hasn't) been achieved. Not sure if this was captured in the EP. If so, disregard. E.g.:
```yaml
conditions:
- type: DisabledCopiedCSVs
  status: True
```

Or:

```yaml
conditions:
- type: DisabledCopiedCSVs
  status: False
  reason: DeleteCSVFailure
  message: 'delete csv "foo" in namespace "bar": not permitted'
```

pkg/operators/v1/olmconfig_types.go
```diff
@@ -0,0 +1,71 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	DisableCopiedCSVsConditionType = "DisableCopiedCSVsConfiguration"
```
joelanford, line 8: Should this be past tense so that it reads a little more like some state has (or hasn't) been achieved? Not sure if this was captured in the EP. If so, disregard. E.g.:
```yaml
conditions:
- type: DisabledCopiedCSVs
  status: True
```

Or:

```yaml
conditions:
- type: DisabledCopiedCSVs
  status: False
  reason: DeleteCSVFailure
  message: 'delete csv "foo" in namespace "bar": not permitted'
```

awgreene: This seems fine and was not captured in the EP

awgreene: Addressed in the latest version of the PR. Note that it isn't possible to discern the state of the condition without both the `status` and `reason`:

`status` is false because copied csvs are enabled:
```yaml
conditions:
- type: DisabledCopiedCSVs
  status: False
  reason: CopiedCSVs enabled
  message: CSVs are being copied
```

vs.

`status` is false because not all copied csvs were deleted:
```yaml
conditions:
- type: DisabledCopiedCSVs
  status: False
  reason: DeleteCSVFailure
  message: 'delete csv "foo" in namespace "bar": not permitted'
```

Not really an issue, but something worth calling out in the eventual docs.

PR #200, opened 2021-12-06
Add format to verify target

awgreene:
This commit updates the verify target to include the
format target as a prerequisite. This commit also
includes changes introduced after running the format
target.


crds/zz_defs.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by go-bindata. (@generated) DO NOT EDIT.
 
- //Package crds generated by go-bindata.// sources:
```
timflannagan, line 3: Do we want to be formatting generated files? Can we adjust the fmt target to filter out certain files/directories/etc.?

estroz: IMO if the generator isn't formatting Go files it has a bug, so I'm fine with formatting everything (with the exception of vendor dirs).

PR #202, opened 2021-12-08
feat(cel): Add CEL custom library for semver comparison

dinhxuanvu:
Add CEL custom library for semver comparison to api repo. This
library is used in olm and registry.


pkg/cel/cel.go
```diff
@@ -0,0 +1,141 @@
...
+	Rule string `json:"rule" yaml:"rule"`
+}
+
+type Evaluator interface {
+	Evaluate(env map[string]interface{}) (bool, error)
+}
+
+type EvaluatorProvider interface {
+	Evaluator(rule string) (Evaluator, error)
+}
```
joelanford, line 40: It would be good to get some GoDoc on these types to make it more obvious what their purpose is.

Are these specific to `cel`? If not, should they be in another package? Or should we name this package something else (e.g. `constraint`)?

pkg/cel/cel.go
```diff
@@ -0,0 +1,141 @@
...
+
+type Constraint struct {
+	// Constraint message that surfaces in resolution
+	// This field is optional
+	Message string `json:"message" yaml:"message"`
+
+	// The cel struct that contraints CEL expression
+	// This field is required
+	Cel *Cel `json:"cel" yaml:"cel"`
+}
```
joelanford, line 27: This probably shouldn't be in the `cel` package right?

joelanford (changes requested):
Quick review!

pkg/cel/cel.go
dinhxuanvu: On the OLM PR, I call this package `constraints` but in api repo, I'm not sure if I should be more explicit about the constraint type. So I call it `cel` here but if `constraints` sounds more appropriate then I'm totally fine wit that.

pkg/cel/cel.go
dinhxuanvu: If we change this package to `constraints` then I think it can be here. See above comment.

pkg/cel/cel.go
joelanford: Yep, if this package is renamed to `constraint` or `constraints`, it would make sense to me to have something like:
```
$ tree constraint
constraint
├── cel.go (cel-specific stuff here)
├── compound.go (All, Any, None constraints here)
└── constraint.go (common types and interfaces, e.g. Constraint and Evaluator*)
```

WDYT?

pkg/cel/cel.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/interpreter/functions"
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+
+	"github.com/blang/semver/v4"
+)
+
+// PropertiesKey is the key for bundle properties map (input data for CEL evaluation)
+const PropertiesKey = "properties"
```
joelanford, line 17: Does this need to be exported?

pkg/cel/cel.go
joelanford: Actually, looking more closely at the evaluator interface, it looks cel-specific (at least for now)?

I'm curious if there's a more generic constraint interface that all of the constraint types could implement? When @estroz goes to add the compound constraint struct, will there be an interface he can implement such that he doesn't have to change anything on the OLM side?

pkg/cel/cel.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func NewCelEvaluatorProvider() *celEvaluatorProvider {
+	env, err := cel.NewEnv(cel.Declarations(
+		decls.NewVar(PropertiesKey, decls.NewListType(decls.NewMapType(decls.String, decls.Any)))),
+		cel.Lib(semverLib{}),
+	)
+	if err != nil {
+		panic(err)
+	}
```
joelanford, line 49: Should we add a comment here explaining why we're panic-ing rather than returning an error?

pkg/cel/cel.go
```diff
@@ -0,0 +1,141 @@
...
+
+type celEvaluator struct {
+	program cel.Program
+}
+
+/*
+This section of code is for custom library for semver comparison in CEL
+The code is inspired by https://github.com/google/cel-go/blob/master/cel/cel_test.go#L46
+
+The semver_compare is wrriten based on `Compare` function in https://github.com/blang/semver
```
perdasilva, line 67:
```suggestion
The semver_compare is written based on `Compare` function in https://github.com/blang/semver
```

pkg/cel/cel_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	propertiesMap := map[string]interface{}{"properties": props}
+
+	tests := []struct {
+		name  string
+		rule  string
+		out   bool
+		isErr bool
+	}{
```
perdasilva, line 23: looks good. I would suggest having a check for each comparison direction in semver to really make sure it works, i.e.
in addition to testing `properties.exists(p, p.type == 'olm.test' && (semver_compare(p.value, '1.0.0') == 0))` add `properties.exists(p, p.type == 'olm.test' && (semver_compare(p.value, '0.9.0') > 0))` and `properties.exists(p, p.type == 'olm.test' && (semver_compare(p.value, '1.0.1') < 0))`

perdasilva: it might also be worth exercising the bad semver code path

perdasilva: also the non-boolean return type error code path

estroz:
Since `olm.constraint` is just a property, which are sent over gRPC automatically, and known only to OLM's resolver, I don't see a need to encode its behavior here for operator-registry's sake. If anything in operator-registry is required to know about constraint implementations (like `opm alpha diff` libraries), then en/decoding for specific implementations should be implemented internally there. If we ever decide to change this split of concerns in the future, then we break no APIs because there is no shared code.

pkg/cel/cel_test.go
dinhxuanvu: Will see what I can do here.

pkg/cel/cel.go
dinhxuanvu: I don't see the need for it right now.

pkg/cel/cel.go
dinhxuanvu: The original code was meant to use with `Evaluator()` function below as a "selector" on which provider to use via `id` parameter (if you remember the original EP with `evaluator` struct with `id` field). Now that we don't have `id` part anyway, this provider becomes CEL-specific which is fine. If we introduce a new "provider" later, we can simply create a new `NewSomethingEvaluatorProvider` and add it to the `SatResolver`. Now `evaluatorProvider` on SatResolver is a field with a single provider but it can be a collection of providers. That shouldn't be a problem code-wise later on.

pkg/cel/cel.go
dinhxuanvu: On a second thought, this is actually used in `predicates.go` on OLM side so it is exported.

dinhxuanvu:
After a discussion with Eric, it seems the generic interface design is not needed yet given only CEL is supported at the moment. There seems to be quite a bit of discussions about this interface that may need more time to figure out. For now, we can simply use struct and functions for CEL evaluation.

pkg/constraints/cel.go
```diff
@@ -0,0 +1,146 @@
...
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+
+	"github.com/blang/semver/v4"
+)
+
+// PropertiesKey is the key for bundle properties map (input data for CEL evaluation)
+const PropertiesKey = "properties"
+
+// Constraint is a struct representing the new generic constraint type
+type Constraint struct {
```
estroz, line 20: Is the intent for `olm.constraint`'s definition to live in the api repo? I was under the impression that OLM's resolver would define this schema.

dinhxuanvu: I think it should be here given registry is gonna use it as well. So this is a shared code. I think Joe wants it here as well. I don't have a strong opinion about this. I can move it back to OLM and registry will have a copy of these 2 structs. It is not a big deal given the code is small.

estroz: Right, I see it will be used [here](https://github.com/operator-framework/operator-registry/pull/879/files#diff-c3b6b7bfc976a561eb2abaf01a8f2de39d5c2ff616bc134daf695ae5d4eccd74R353) at least. SGTM.

estroz:
I'm fine with how this PR stands now. We can always deprecate the types defined here once we have a more robust interface system designed to handle other constraint types as @joelanford has suggested [here](https://github.com/operator-framework/api/pull/202#discussion_r765220031).

dinhxuanvu:
I'm putting a hold here so people can approve and lgtm.
/hold

dinhxuanvu:
Ready for merge
/hold cancel

PR #203, opened 2021-12-10
feat(constraints): add compound constraints and olm.constraint value parser

estroz:
Follow-up on #202 for https://github.com/operator-framework/enhancements/blob/master/enhancements/compound-bundle-constraints.md

### DIFF
--- a/pkg/constraints/cel.go
+++ b/pkg/constraints/cel.go
@@ -16,17 +16,6 @@ import (
 // PropertiesKey is the key for bundle properties map (input data for CEL evaluation)
 const PropertiesKey = "properties"
 
-// Constraint is a struct representing the new generic constraint type
-type Constraint struct {
-	// Constraint message that surfaces in resolution
-	// This field is optional
-	Message string `json:"message" yaml:"message"`
-
-	// The cel struct that contraints CEL expression
-	// This field is required
-	Cel *Cel `json:"cel" yaml:"cel"`
-}
-
 // Cel is a struct representing CEL expression information
 type Cel struct {
 	// The CEL expression
--- a/pkg/constraints/constraint.go
+++ b/pkg/constraints/constraint.go
@@ -38,22 +38,22 @@ type Constraint struct {
 // CompoundConstraint holds a list of potentially nested constraints
 // over which a boolean operation is applied.
 type CompoundConstraint struct {
-	Constraints []Constraint `json:"constraints"`
+	Constraints []Constraint `json:"constraints" yaml:"constraints"`
 }
 
 // GVKConstraint defines a GVK constraint.
 type GVKConstraint struct {
-	Group   string `json:"group"`
-	Kind    string `json:"kind"`
-	Version string `json:"version"`
+	Group   string `json:"group" yaml:"group"`
+	Kind    string `json:"kind" yaml:"kind"`
+	Version string `json:"version" yaml:"version"`
 }
 
 // PackageConstraint defines a package constraint.
 type PackageConstraint struct {
-	// Name of the package.
-	Name string `json:"name"`
+	// PackageName is the name of the package.
+	PackageName string `json:"packageName" yaml:"packageName"`
 	// VersionRange required for the package.
-	VersionRange string `json:"versionRange"`
+	VersionRange string `json:"versionRange" yaml:"versionRange"`
 }
 
 // maxConstraintSize defines the maximum raw size in bytes of an olm.constraint.
--- a/pkg/constraints/constraint.go
+++ b/pkg/constraints/constraint.go
@@ -1,0 +1,86 @@
+package constraints
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// OLMConstraintType is the schema "type" key for all constraints known to OLM
+// (except for legacy types).
+const OLMConstraintType = "olm.constraint"
+
+// Constraint holds parsed, potentially nested dependency constraints.
+type Constraint struct {
+	// Constraint message that surfaces in resolution
+	// This field is optional
+	Message string `json:"message,omitempty" yaml:"message,omitempty"`
+
+	// The cel struct that contraints CEL expression
+	Cel *Cel `json:"cel,omitempty" yaml:"cel,omitempty"`
+
+	// Package defines a constraint for a package within a version range.
+	Package *PackageConstraint `json:"package,omitempty" yaml:"package,omitempty"`
+
+	// GVK defines a constraint for a GVK.
+	GVK *GVKConstraint `json:"gvk,omitempty" yaml:"gvk,omitempty"`
+
+	// All, Any, and None are compound constraints. See this enhancement for details:
+	// https://github.com/operator-framework/enhancements/blob/master/enhancements/compound-bundle-constraints.md
+	All *CompoundConstraint `json:"all,omitempty" yaml:"all,omitempty"`
+	Any *CompoundConstraint `json:"any,omitempty" yaml:"any,omitempty"`
+	// A note on None: this constraint is not particularly useful by itself.
+	// It should be used within an All constraint alongside some other constraint type
+	// since saying "none of these GVKs/packages/etc." without an alternative doesn't make sense.
+	None *CompoundConstraint `json:"none,omitempty" yaml:"none,omitempty"`
+}
+
+// CompoundConstraint holds a list of potentially nested constraints
+// over which a boolean operation is applied.
+type CompoundConstraint struct {
+	Constraints []Constraint `json:"constraints"`
+}
+
+// GVKConstraint defines a GVK constraint.
+type GVKConstraint struct {
+	Group   string `json:"group"`
+	Kind    string `json:"kind"`
+	Version string `json:"version"`
+}
+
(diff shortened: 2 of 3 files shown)


pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+type GVKConstraint struct {
+	Group   string `json:"group"`
+	Kind    string `json:"kind"`
+	Version string `json:"version"`
+}
+
+// PackageConstraint defines a package constraint.
+type PackageConstraint struct {
+	// Name of the package.
+	Name string `json:"name"`
```
dinhxuanvu, line 53: I'm not sure if the package constraint has changed but this doesn't look correct to me. It is supposed to be `packageName` and `Version`? Do we unmarshal the JSON blob directly into this `PackageConstraint` struct or this is struct is some sort of internal representation?

pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+	// A note on None: this constraint is not particularly useful by itself.
+	// It should be used within an All constraint alongside some other constraint type
+	// since saying "none of these GVKs/packages/etc." without an alternative doesn't make sense.
+	None *CompoundConstraint `json:"none,omitempty" yaml:"none,omitempty"`
+}
+
+// CompoundConstraint holds a list of potentially nested constraints
+// over which a boolean operation is applied.
+type CompoundConstraint struct {
+	Constraints []Constraint `json:"constraints"`
```
dinhxuanvu, line 40: While I understand `CompoundConstraint` is a list of constraints but then `Constraint` itself can be a `All`, `Any`, `None` constraint which is a `CompoundConstraint` itself. So we potentially have a `CompoundConstraint` as a list of `CompoundConstraint`. While it is not a problem programmatically but logically it is convoluted and frankly difficult to test.

pkg/constraints/constraint.go
estroz: It was `packageName` on `olm.package.required` but this is a completely different type so the name doesn't have to match. `packageName` is redundant anyway.

pkg/constraints/constraint.go
estroz: Not quite, since `CompoundConstraint` does not alias `Constraint`; they're two distinct types. I don't see how you would get nice recursion that leverages std serialization libs otherwise. The [encoded form](https://github.com/operator-framework/operator-lifecycle-manager/pull/2418/files#diff-ea16582ca96e853d17883479654c218897cb4bf126ddd3c0d4908d24ef99329eR806-R812) is readable and comprehensible imo. Plus this is what the EP proposed.

pkg/constraints/constraint.go
dinhxuanvu: I'm slightly confused. So this `PackageConstraint` is a new type of constraint? Or you are saying that if you want to use compound constraint then you need to use this new `PackageConstraint`, not the existing `olm.package.required`?

estroz: If you want to use `olm.constraint` then you need to use the new `PackageConstraint` type. The only difference for users is `packageName -> name` (other than the different parent schema).

pkg/constraints/cel.go
```diff
@@ -16,17 +16,6 @@ import (
...
-type Constraint struct {
-	// Constraint message that surfaces in resolution
-	// This field is optional
-	Message string `json:"message" yaml:"message"`
-
-	// The cel struct that contraints CEL expression
-	// This field is required
-	Cel *Cel `json:"cel" yaml:"cel"`
-}
-
```
awgreene, line 29: Note: This was just moved to `constraints.go` and is still available within the package.

pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+
+// OLMConstraintType is the schema "type" key for all constraints known to OLM
+// (except for legacy types).
+const OLMConstraintType = "olm.constraint"
+
+// Constraint holds parsed, potentially nested dependency constraints.
+type Constraint struct {
+	// Constraint message that surfaces in resolution
+	// This field is optional
+	Message string `json:"message"`
```
awgreene, line 17: Every other field has both yaml and json tags defined, should this field include a yaml tag? Applies everywhere.

pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// The cel struct that contraints CEL expression
+	Cel *Cel `json:"cel,omitempty" yaml:"cel,omitempty"`
+
+	// Package defines a constraint for a package within a version range.
+	Package *PackageConstraint `json:"package,omitempty" yaml:"package,omitempty"`
+	// GVK defines a constraint for a GVK.
```
awgreene, line 24: Non-blocking Nit: Newline.

pkg/constraints/constraint.go
awgreene: Just to confirm what you were saying @estroz, the `olm.constraint.package` is a more fleshed out alternative to using `olm.package.required` and `olm.package.required` will not be unsupported with this change.

pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+	Name string `json:"name"`
+	// VersionRange required for the package.
+	VersionRange string `json:"versionRange"`
+}
+
+// maxConstraintSize defines the maximum raw size in bytes of an olm.constraint.
+// 64Kb seems reasonable, since this number allows for long description strings
+// and either few deep nestings or shallow nestings and long constraints lists,
+// but not both.
+const maxConstraintSize = 2 << 16
```
awgreene, line 62: Non-blocking: I am weary of hardcoded maximums that users cannot override.

pkg/constraints/constraint.go
awgreene: This is user configurable eh? Slightly concerned what could be written here, but it should be fine.

pkg/constraints/constraint_test.go
```diff
@@ -0,0 +1,274 @@
...
+		},
+		{
+			name:  "Valid/BasicAll",
+			input: json.RawMessage(fmt.Sprintf(inputBasicCompoundTmpl, "all")),
+			expConstraint: Constraint{
+				Message: "blah",
+				All: &CompoundConstraint{
+					Constraints: []Constraint{
+						{
+							Message: "blah blah",
```
awgreene, line 45: Note: We have moved on from "blah".

pkg/constraints/constraint.go
```diff
@@ -0,0 +1,83 @@
...
+
+// ErrMaxConstraintSizeExceeded is returned when a constraint's size > maxConstraintSize.
+var ErrMaxConstraintSizeExceeded = fmt.Errorf("olm.constraint value is greater than max constraint size %d bytes", maxConstraintSize)
+
+// Parse parses an olm.constraint property's value recursively into a Constraint.
+// Unknown value schemas result in an error. A too-large value results in an error.
```
awgreene, line 68:
```suggestion
// Unknown value schemas result in an error. Constraints that exceed the number of bytes defined by maxConstraintSize result in an error.
```

awgreene (commented):
Looks good to me. Primarily non-blocking nits and generic comments.

pkg/constraints/constraint.go
estroz: Huh not sure where the yaml tag went, will re-add.

pkg/constraints/constraint.go
estroz: This is for security purposes so no one should be able to configure this, at least not for now. I'll add a note to come back to this later.

pkg/constraints/constraint.go
dinhxuanvu: So if they want to use solo package dependency, they need to use existing schema but if they want to use compound, they need to use the new scheme even though they are the same but just different field names. I’m not entirely sure what the purpose of this is? It is an unnecessary introduction of a new schema for the sake of compound constraints even they are basically the same. If we continue to use the existing schema for compound it should still work.

estroz: Yep!

estroz: I'm not tied to name over packageName, but that is [what is approved](https://github.com/operator-framework/enhancements/blob/master/enhancements/compound-bundle-constraints.md#implementation-detailsnotesconstraints) in the EP.

awgreene:
/approve

I'm happy with the changes introduced in this commit, but I will wait for @dinhxuanvu to lgtm as he has been involved in the PR.

pkg/constraints/constraint_test.go
```diff
@@ -0,0 +1,274 @@
...
+			},
+		},
+		{
+			name: "Invalid/TooLarge",
+			input: func(t *testing.T) json.RawMessage {
+				p := make([]byte, maxConstraintSize+1)
+				_, err := rand.Read(p)
+				require.NoError(t, err)
+				return json.RawMessage(p)
+			}(t),
```
timflannagan, line 129: Nice!

dinhxuanvu (approved):
There is a bit about version vs versionRange for package dependency that I need to look into but lgtm overall.
/lgtm

PR #204, opened 2021-12-10
bugfix: make spec.grpcPodConfig.priorityClassName a *string

perdasilva:
### Description
This PR updates the CatalogSource spec.grpcPodConfig.priorityClassName to be a *string. The main motivation being that empty string has meaning for this attribute so we need to distinguish whether it was set by the user or not.


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -107,7 +107,7 @@ type GrpcPodConfig struct {
 	// If not specified, the pod priority will be default or zero if there is no
 	// default.
 	// +optional
-	PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,24,opt,name=priorityClassName"`
+	PriorityClassName *string `json:"priorityClassName,omitempty" protobuf:"bytes,24,opt,name=priorityClassName"`
```
estroz, line 110: Where is this protobuf tag coming from? Can this be removed?

estroz:
What's the difference between `""` and unset?

Overall I think this is fine because most consumers define their CatalogSource's via yaml configs, so this doesn't affect most people.

pkg/operators/v1alpha1/catalogsource_types.go
perdasilva: oh nice catch! thank you!!!

perdasilva:
if they leave it unset it, nothing will happen. If they set it to "", then we'll set it to "" in the pod, basically. I'll fix the protobuf tags!!

pkg/operators/v1alpha1/catalogsource_types.go
perdasilva: Done!

estroz:
Sounds good. Lets see if I can get prow to override go-apidiff

/override Go/go-apidiff
/approve
/lgtm

estroz:
Merging by hand

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -2,12 +2,11 @@ package v1alpha1
...
 import (
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
-	"time"
-
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"time"
```
timflannagan, line 9: Interesting - I'm surprised the format check didn't complain about this but I guess gofmt might no care about import ordering/chunking/etc.?

estroz: It doesn't have an opinion about std vs. external module chunking, just sorts within chunks. goimports does care i think.

PR #209, opened 2022-01-06
Added field for optional manifests

fgiloux:
This pull request adds a new Optional field to the Step structure. This is to allow successful InstallPlan processing when an optional manifest can't be created. See [the optional manifest enhancement proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/optional-manifest.md)

Signed-off-by: Frederic Giloux <fgiloux@redhat.com>


pkg/operators/v1alpha1/installplan_types.go
```diff
@@ -227,6 +228,7 @@ func ConditionMet(cond InstallPlanConditionType, now *metav1.Time) InstallPlanCo
 type Step struct {
 	Resolving string       `json:"resolving"`
 	Resource  StepResource `json:"resource"`
+	Optional  bool         `json:"optional,omitempty"`
```
timflannagan, line 231: Quick question: I haven't read through the original enhancement entirely so apologies if has already been addressed - does this need to be a pointer to a bool?

timflannagan:
This lgtm outside of the question around whether the `Optional` struct field needs to be a `*bool`.

pkg/operators/v1alpha1/installplan_types.go
fgiloux: @timflannagan Thanks for looking at it. It is indeed the way it was specified in the[ enhancement proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/optional-manifest.md#step). That said it is possible to amend it if there is a case for it.
I understand that a bool is not extensible (it accepts two values and that's it). Originally I had played with the idea of having a field named Criticality instead of Optional, accepting values like: "Mandatory", "Optional" and possibly others. The thing is that I was advised to have the list of optional manifests in a bundle property rather than having the criticality specified in an annotation directly in the manifest. As an entry in a list the file is either there or not. This is binary... like a boolean. Keeping the same interface I don't see any value in making it a string but I am open to read your views.

timflannagan: Yeah definitely - IIRC, the need for a pointer to a bool in relevant when we need to distinguish between a user supplied value (i.e. it was explicitly set) vs. the default/empty value from Go in the context of boolean types. If that isn't the case here, then we should be all set to settle on a bool.

timflannagan:
I think I've convinced myself a pointer to a bool isn't necessary, and we can always cut a new patch release is we discover we need to handle this differently once the OLM implementation is fleshed out.

/lgtm

pkg/operators/v1alpha1/installplan_types.go
fgiloux: ok. Thanks for clarifying I had missed the pointer aspect.

timflannagan:
@fgiloux It looks like you need to regenerate the CRDs as the verify check is failing. Try running `make manifests` locally and seeing if there's a diff that's produced.

dinhxuanvu (approved):
/approve

PR #210, opened 2022-01-08
add bundle size validator check

camilamacedo86:
**Description**
- add bundle size validator check
- fix warn level that was as error type

Co-authored-by: Brett Tofel btofel@redhat.com

Closes: https://github.com/operator-framework/api/pull/163

Related to: https://github.com/operator-framework/operator-registry/pull/685


pkg/validation/internal/bundle.go
```diff
@@ -15,6 +17,9 @@ import (
 
 var BundleValidator interfaces.Validator = interfaces.ValidatorFunc(validateBundles)
 
+// max_bundle_sized_allowed defines the max size of the bundle
+const max_bundle_sized_allowed = 1048576
```
camilamacedo86, line 21: @kevinrizza could you please help out just by double-checking that is the correct amount?

camilamacedo86:
c/c @bentito @gallettilance @perdasilva

pkg/validation/internal/bundle.go
camilamacedo86: c/c @perdasilva @njhale @anik120 could you please help us check if that amount is the right one and etc?

njhale (changes requested):

pkg/validation/internal/bundle.go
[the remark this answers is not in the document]
camilamacedo86: So, do you think that we can safely increase this amount to ~4MB? 
I know that currently if the bundle is too big that fails in the deploy right? 
The validator here would give the opportunity we alert and give a proper message before that happens.

pkg/validation/internal/bundle.go
```diff
@@ -117,6 +126,30 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
...
 
+// validateBundleSize will raise an error if the bundle size is bigger than the value allowed
+// or a warning when the value is closed to it
+func validateBundleSize(bundle *manifests.Bundle) []errors.Error {
+	warnPercent := 0.15
+	warnSize := int64(max_bundle_sized_allowed - math.Round(max_bundle_sized_allowed*warnPercent))
+	var errs []errors.Error
+
+	if bundle.Size == nil || *bundle.Size == 0 {
+		errs = append(errs, errors.WarnFailedValidation("unable to check the bundle size", nil))
```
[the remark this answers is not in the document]
camilamacedo86: Should we really raise an error in this case?
It shows safer a warning because we do not want getting started to block/fail in validation because in some edge cases where the loader was not called.

kevinrizza:
Is it also worth checking each file to ensure that they can be applied by etcd? Even if OLM can unpack it, if a single crd hits the default etcd file limit on clusters (1MB), then the bundle won't be applied just a little further into install

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-bindata/go-bindata/v3 v3.1.3
 	github.com/google/cel-go v0.9.0
 	github.com/mikefarah/yq/v3 v3.0.0-20201202084205-8846255d1c37
+	github.com/operator-framework/operator-registry v1.19.5
```
kevinrizza, line 12: this is a circular dependency. the api project shouldn't depend on the operator-registry

camilamacedo86: So, it shows for me that the code to do the zip ought to be shipped here and we do a follow up for opm get it from here. Tks
/hold

camilamacedo86: HI @kevinrizza, 

That is changed. 

/hold cancel

jmrodri:
@camilamacedo86 rename this file: `pkg/encoding/enconding.go` I think it should be `encoding.go`

jmrodri (commented):
/lgtm

jmrodri (approved):
/lgtm

pkg/validation/internal/bundle.go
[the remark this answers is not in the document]
njhale: I think so

njhale (approved):
/approve

PR #212, opened 2022-01-12
chore: Bump k8s dependencies to 1.23

varshaprasad96:
Some notable bumps in versions:
1. Controller-runtime to 0.11.0
2. Controller-tools to 0.8.0
3. K8s modules to 1.23+
4. Go to 1.17


joelanford:
Unrelated to this PR, but is there anything blocking us from removing the `vendor` directory from this repo?

/cc @timflannagan @dinhxuanvu @perdasilva

timflannagan:
+1 to removing the vendor directory from this repository. I don't think it makes a ton of sense upstream, and would simplify the downstreaming process (e.g. we run into less merge conflicts). Do we typically bump this repositories' minor version tag when we cut new k8s dependencies? If so, it might be worth trying to see whether purging the vendor directory is a small amount of work and piggy back off that new minor version.

dinhxuanvu:
I have no problems with removing vendor as long as things continue to work the way they are (in CI and etc).

timflannagan:
> Unrelated to this PR, but is there anything blocking us from removing the vendor directory from this repo?

Took a stab at removing the vendor directory in https://github.com/operator-framework/api/pull/213. It would real nice to also remove it from the registry repository as well (we're tracking the OLM removal as an upstream issue, but it's a bit more nuanced to remove it than I'd like).

.github/workflows/go.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16
```
timflannagan, line 17: super nit (feel free to ignore): It would be nice to separate these types of changes into their own commit so viewing the overall diff outside of vendor changes doesn't result in a decent lag in the github UI.

camilamacedo86 (approved):
/approved

dinhxuanvu (approved):
Approved.

PR #213, opened 2022-01-13
Purge the root vendor directory

timflannagan:
Update the repository layout and avoid commiting the root vendor directory into source. Update the root .gitignore and ignore the vendor directory. For any CI/CD (or utility-esq Makefile targets) avoid referencing the vendor equivalent through `go run ...` and instead use the go-get-tool from the operator-sdk/kubebuilder world) and pull down that dependency at runtime.


timflannagan:
I realize we need to also bump this repository to Go 1.17.

tools.go
```diff
@@ -3,10 +3,6 @@
 package tools
 
 import (
-	// Generate deepcopy and conversion.
-	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"
-	// Manipulate YAML.
-	_ "github.com/mikefarah/yq/v3"
 	// Generate embedded files.
 	_ "github.com/go-bindata/go-bindata/v3/go-bindata"
```
joelanford, line 7: Are we able to use go-get-tool for go-bindata as well?

Makefile
```diff
@@ -88,9 +86,28 @@ test-unit: ## Run the unit tests
...
+# go-get-tool will 'go get' any package $2 and install it to $1.
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-get-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(PROJECT_DIR)/bin go get $(2) ;\
```
joelanford, line 102: Just as a heads up.. this is fine for now, but it's going to break when we bump to Go 1.18. See https://github.com/operator-framework/operator-sdk/issues/5481

timflannagan: That's a good call out - I'm personally not super worried if we're committed to tracking this here as an issue (or ideally in jira ticket) as it looks like a fairly straight forward change, and removing the vendor directory is a step in the right directory (both in the context of upstream, and especially downstream) imo.

timflannagan:
I'd like to resolve that go-apidiff before merging this to avoid putting this repository in a state where we break that action, and potentially evaluate purging the root tools.go file.

/hold

timflannagan:
Ah, it looks like we don't have the workflow_dispatch event trigger enabled for this repository. Opened a quick PR (https://github.com/operator-framework/api/pull/214) for enabling that for the two workflow definitions.

tools.go
timflannagan: Surprisingly, this was a bit more complicated that I had anticipated due to how bloated the `manifests` target is right now, and exporting the PATH variable to point to this repositories' root `./bin` directory before running `go generate ...` wasn't working, and that command was always searching for packages in the vendor directory, despite no GOFLAGS being set in my local environment.

It's possible I'm missing something obvious, but given the o-f/api recent changes that contain massive changelogs due to dependency bumps, maybe we can tackle this in a follow-up, and/or generalize the value we see in the tools.go file going forward throughout the OLM-controller repositories?

fgiloux:
I have seen how annoying the vendor directory is during PR. That said there are good reasons for keeping a copy of the dependencies. A possible mitigation could be to have the vendor directory stored in a separate repository when a new release is cut. What do you think?

timflannagan:
I think the value of the vendor directory has somewhat deteriorated as modules become more stable, and my preference would be to remove the vendor directory from upstream repositories, while maintaining the vendor directory in any downstream environments. That should serve as a nice balance where we can rely on dependencies being available for downstream builds, but help improve the upstream contributor/reviewer experience. Thoughts?

fgiloux:
> to remove the vendor directory from upstream repositories, while maintaining the vendor directory in any downstream 

Yes that sounds good to me. My main concern was indeed when there is a need to patch a downstream release and the dependency version is nowhere to be found.

dinhxuanvu (approved):
/approve

Makefile
timflannagan: I'll create an ticket in this repository once this PR merges to track this.

PR #215, opened 2022-01-15
fix: alm-example parse check should not fail when the annotation if empty

camilamacedo86:
**Description**
Fix error found using the check to test the images data.
Error `"Error: Value invalid character at 0\n \u003c--(see the invalid character): invalid example"` is faced when the alm-example is empty and should not.

Follow up: https://github.com/operator-framework/api/pull/207

PR #217, opened 2022-01-18
feat: add warning for olm.properties being defined when validating csv

tylerslaton:
## Summary
To sufficiently address this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2018095), this PR adds a warning when a user has defined the same property in the CSV's olm.properties annotation as well as in a properties.yaml. This will act as a soft barrier to users doing this in the future.

## Why this is needed
As the `olm.properties` annotation currently exists to just maintain backwards compatibility, we should warn newly created bundles from using it. Instead, it is a good idea to push them to use the `metadata/properties.yaml` file.

### DIFF
--- a/pkg/validation/internal/community.go
+++ b/pkg/validation/internal/community.go
@@ -3,10 +3,11 @@ package internal
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/blang/semver"
 	"io/ioutil"
 	"os"
 	"strings"
+
+	"github.com/blang/semver"
 
 	"github.com/operator-framework/api/pkg/manifests"
 	"github.com/operator-framework/api/pkg/validation/errors"
--- a/pkg/validation/internal/testdata/correct.csv.olm.properties.annotation.yaml
+++ b/pkg/validation/internal/testdata/correct.csv.olm.properties.annotation.yaml
@@ -1,0 +1,60 @@
+#! validate-crd: deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
+#! parse-kind: ClusterServiceVersion
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: etcdoperator.v0.9.0
+  namespace: placeholder
+  annotations:
+    alm-examples: '[{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdCluster","metadata":{"name":"example","namespace":"default"},"spec":{"size":3,"version":"3.2.13"}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdRestore","metadata":{"name":"example-etcd-cluster"},"spec":{"etcdCluster":{"name":"example-etcd-cluster"},"backupStorageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdBackup","metadata":{"name":"example-etcd-cluster-backup"},"spec":{"etcdEndpoints":["<etcd-cluster-endpoints>"],"storageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}}]'
+    olm.properties: '[{"type": "foo", "value": "bar"}]'
+spec:
+  version: 0.9.0
+  installModes:
+  - type: AllNamespaces
+    supported: true
+  install:
+    strategy: deployment
+    spec:
+      permissions:
+      - serviceAccountName: etcd-operator
+        rules:
+        - apiGroups:
+          - etcd.database.coreos.com
+          resources:
+          - etcdclusters
+          - etcdbackups
+          - etcdrestores
+          verbs:
+          - "*"
+      deployments:
+      - name: etcd-operator
+        spec:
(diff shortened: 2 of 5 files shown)


pkg/validation/internal/community.go
```diff
@@ -24,6 +24,9 @@ const ocpLabelindex = "com.redhat.openshift.versions"
 // OCP version where the apis v1beta1 is no longer supported
 const ocpVerV1beta1Unsupported = "4.9"
 
+// olm.properties is a CSV annotation that defines the properties of a bundle.
+const olmproperties = "olm.properties"
```
kevinrizza, line 28: this seems like an unrelated gofmt change? does this need to be included in the commit?

tylerslaton: So this was actually done by me so that I could run this if statement against a constant, instead.

```
if annotationKey == olmproperties {
```

ankitathomas (approved):
/lgtm

pkg/validation/internal/testdata/correct.csv.olm.properties.annotation.yaml
```diff
@@ -0,0 +1,293 @@
...
+metadata:
+  name: etcdoperator.v0.9.0
+  namespace: placeholder
+  annotations:
+    olm.properties: '[{"type": "foo", "value": "bar"}]'
+    capabilities: Full Lifecycle
+    tectonic-visibility: ocs
+    alm-examples: '[{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdCluster","metadata":{"name":"example","namespace":"default"},"spec":{"size":3,"version":"3.2.13"}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdRestore","metadata":{"name":"example-etcd-cluster"},"spec":{"etcdCluster":{"name":"example-etcd-cluster"},"backupStorageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdBackup","metadata":{"name":"example-etcd-cluster-backup"},"spec":{"etcdEndpoints":["<etcd-cluster-endpoints>"],"storageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}}]'
+    description: etcd is a distributed key value store providing a reliable way to store data across a cluster of machines.
+spec:
```
dinhxuanvu, line 14: Would you mind cleaning up this CSV a bit AKA stripping up all unnecessary content here to keep its size as small as possible? Look at this [PR](https://github.com/operator-framework/api/pull/215/files) as an example.

dinhxuanvu (changes requested):
Looking good. Just one small request.

pkg/validation/internal/testdata/correct.csv.olm.properties.annotation.yaml
tylerslaton: Sure, that's a good call. Let me do that now.

dinhxuanvu (approved):
/approve

pkg/validation/internal/community.go
camilamacedo86, line 28: @tylerslaton this validator is deprecated we must do these changes in the https://github.com/redhat-openshift-ecosystem/ocp-olm-catalog-validator. This implementation was moved to this project because the max OCP version is OCP specific. 

We have tasks already asking for the pipelines and CI's do the replacement. Could you please give a hand here and push this one for the right project?

camilamacedo86 (changes requested):
/hold

1) Could we not address this one by:

a) In the CSV validator check if properties are define and then yes warning asking to create the file properties instead? 
b) In the bundle validator check if the CSV and the properties file are defining the same property? 

After some thoughts about this scenario, I think that it should not be specific to the max ocp version and we could generalize better this one. Also, by doing the above suggestions seems that we would be able to properly solve the BZ by not allowing any duplications.

pkg/validation/internal/community.go
camilamacedo86: It shows outdate based on; https://github.com/operator-framework/api/pull/217#pullrequestreview-860307771

tylerslaton:
> /hold
> 
> 1. Could we not address this one by:
> 
> a) In the CSV validator check if properties are defined and then if yes warning asking to use the file properties instead? b) In the bundle validator check if the CSV and the properties file are defining the same property? If yes, then raise an error because the bundle is defining the same value twice.
> 
> After some thoughts about this scenario, I think that it should not be specific to the max OCP version and we could generalize better this one. Also, by doing the above suggestions seems that we would be able to properly solve the BZ by not allowing any duplications.

Hi Camilla, so a couple of things.

1. I believe that this PR already satisfies your first point. If you define `olm.properties` in the CSV's annotation then a warning will be thrown asking the user to define it in `metadata/properties.yaml` instead.
2. With this ^ in mind, would it make sense to get this PR merged and then address your second point?

pkg/validation/internal/community.go
tylerslaton: I just went ahead and reverted this change and added it into the CSV validator instead. This should avoid any changes to this deprecated file.

camilamacedo86:
/hold cancel

@tylerslaton 
You are right
/lgtm
/approved

PR #223, opened 2022-01-26
fix(constraints): bring back the none constraint and add a not alias for it

tylerslaton:
Signed-off-by: Tyler Slaton <tyslaton@redhat.com>

# Summary
A mistake was made where the None property was removed instead of just adding an alias for it. This PR adds back None, makes Not the default, and leaves "deprecated" comment on the None property.

### DIFF
--- a/pkg/constraints/constraint.go
+++ b/pkg/constraints/constraint.go
@@ -29,11 +29,15 @@ type Constraint struct {
 	// https://github.com/operator-framework/enhancements/blob/master/enhancements/compound-bundle-constraints.md
 	All *CompoundConstraint `json:"all,omitempty" yaml:"all,omitempty"`
 	Any *CompoundConstraint `json:"any,omitempty" yaml:"any,omitempty"`
-	// A note on Not: this constraint isn't particularly useful by itself.
+	// A note on Not: this constraint is not particularly useful by itself.
 	// It should be used within an All constraint alongside some other constraint type
 	// since saying "do not use any of these GVKs/packages/etc." without an alternative
-	// doesn't make sense.
+	// doesn't make sense. Both Not and None can be used for this type but None is
+	// deprecated. If values for both Not and None are defined, None will be ignored
+	// as Not is preferred.
 	Not *CompoundConstraint `json:"not,omitempty" yaml:"not,omitempty"`
+	// Deprecated: Use Not instead. This has been deprecated as Not has a clearer UX.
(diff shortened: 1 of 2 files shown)


awgreene (approved):
/approve

pkg/constraints/constraint.go
```diff
@@ -29,11 +29,14 @@ type Constraint struct {
...
 	Any *CompoundConstraint `json:"any,omitempty" yaml:"any,omitempty"`
-	// A note on Not: this constraint isn't particularly useful by itself.
+	// A note on Not: this constraint is not particularly useful by itself.
 	// It should be used within an All constraint alongside some other constraint type
 	// since saying "do not use any of these GVKs/packages/etc." without an alternative
-	// doesn't make sense.
+	// doesn't make sense. Both Not and None can be used for this type but None is
+	// deprecated.
 	Not *CompoundConstraint `json:"not,omitempty" yaml:"not,omitempty"`
+	// Deprecated: Use Not instead. This has been deprecated as Not has a clearer UX.
```
joelanford, line 38: What happens if both `Not` and `None` are defined simultaneously? Seems like there are a few options:
1. Ignore `None`
2. Combine `None` and `Not`
3. Error

I'd lean toward 2 or 3 I think. Whatever we choose, it would be nice to add that to the GoDoc.

tylerslaton: What would combining them both together look like to you?

njhale: I vote for option 1. I think adding a comment to both fields with something like "if the values of both `None` and `Not` are defined, `Not` is preferred" would be sufficient to codify the desired API behavior (since we can't always control the marshaller being used by a client).

njhale, line 39: nit: go-doc comments on exported fields must start with the field name:

```go
// None is a...
// Deprecated: Use Not instead...
None *CompoundConstraint `json:...`
```

tylerslaton: Added a comment for this in the most recent push.

njhale: Now that I see the `Parse` method again, I think it should handle pruning things down, but should also return a real error when more than one is defined so that the invoking code can bubble a warning if necessary.

tylerslaton: Should this also be updated for the other fields?

tylerslaton:
Closing as a different approach has been thought of. Going to leave this as is and create a new minor version (0.12.0).

PR #229, opened 2022-02-24
bundle size validator: fix and improvements

camilamacedo86:
- fix: bundle size validator: amount compressed scenarios should be lower than < ~1MB **(required and blocks SDK release 1.18.0)**
- add total uncompressed info and add unit format method to have a better output to help the users

### DIFF
--- a/pkg/manifests/bundle.go
+++ b/pkg/manifests/bundle.go
@@ -20,7 +20,9 @@ type Bundle struct {
 	V1CRDs         []*apiextensionsv1.CustomResourceDefinition
 	Dependencies   []*Dependency
 	// CompressedSize stores the gzip size of the bundle
-	CompressedSize *int64
+	CompressedSize int64
+	// Size stores the size of the bundle
+	Size int64
 }
 
 func (b *Bundle) ObjectsToValidate() []interface{} {
--- a/pkg/manifests/bundleloader.go
+++ b/pkg/manifests/bundleloader.go
@@ -49,26 +49,30 @@ func (b *bundleLoader) LoadBundle() error {
 
 // Compress the bundle to check its size
 func (b *bundleLoader) calculateCompressedBundleSize() error {
+	if b.bundle == nil {
+		return nil
+	}
 	err := filepath.Walk(b.dir,
 		func(path string, info os.FileInfo, err error) error {
 			if err != nil {
 				return err
 			}
-
-			if !info.IsDir() {
-				if data, err := os.ReadFile(path); err == nil {
-					content, err := encoding.GzipBase64Encode(data)
-					if err != nil {
-						return err
-					}
-					total := int64(len(content))
-					if b.bundle.CompressedSize != nil {
-						total += *b.bundle.CompressedSize
-					}
-					b.bundle.CompressedSize = &total
-				}
+			if info.IsDir() {
+				return nil
 			}
-			return nil
+			data, err := os.ReadFile(path)
+			if err == nil {
+				// Sum the bundle amount
+				b.bundle.Size += info.Size()
+
+				// Sum the compressed amount
+				contentGzip, err := encoding.GzipBase64Encode(data)
+				if err != nil {
+					return err
+				}
+				b.bundle.CompressedSize += int64(len(contentGzip))
+			}
+			return err
 		})
 	if err != nil {
 		return err
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -17,8 +17,9 @@ var BundleValidator interfaces.Validator = interfaces.ValidatorFunc(validateBund
 
 // max_bundle_size is the maximum size of a bundle in bytes.
 // This ensures the bundle can be staged in a single ConfigMap by OLM during installation.
-// The value is derived from the standard upper bound for k8s resources (~4MB).
-const max_bundle_size = 4 << (10 * 2)
+// The value is derived from the standard upper bound for k8s resources (~1MB).
+// We will use this value to check the bundle compressed is < ~1MB
+const max_bundle_size = int64(1 << (10 * 2))
 
 func validateBundles(objs ...interface{}) (results []errors.ManifestResult) {
 	for _, obj := range objs {
@@ -133,21 +134,53 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
 // - we could identify that the bundle size is close to the limit (bigger than 85%)
 func validateBundleSize(bundle *manifests.Bundle) []errors.Error {
 	warnPercent := 0.85
-	warnSize := int64(max_bundle_size * warnPercent)
+	warnSize := float64(max_bundle_size) * warnPercent
 	var errs []errors.Error
 
-	if bundle.CompressedSize == nil || *bundle.CompressedSize == 0 {
-		errs = append(errs, errors.WarnFailedValidation("unable to check the bundle size", nil))
+	if bundle.CompressedSize == 0 {
+		errs = append(errs, errors.WarnFailedValidation("unable to check the bundle compressed size", bundle.Name))
 		return errs
 	}
 
-	if *bundle.CompressedSize > max_bundle_size {
-		errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("maximum bundle compressed size with gzip size exceeded: size=~%d MegaByte, max=%d MegaByte", *bundle.CompressedSize/(1<<(10*2)), max_bundle_size/(1<<(10*2))), nil))
-	} else if *bundle.CompressedSize > warnSize {
-		errs = append(errs, errors.WarnInvalidBundle(fmt.Sprintf("nearing maximum bundle compressed size with gzip: size=~%d MegaByte, max=%d MegaByte", *bundle.CompressedSize/(1<<(10*2)), max_bundle_size/(1<<(10*2))), nil))
+	if bundle.Size == 0 {
+		errs = append(errs, errors.WarnFailedValidation("unable to check the bundle size", bundle.Name))
+		return errs
+	}
+
+	// From OPM (https://github.com/operator-framework/operator-registry) 1.17.5
+	// and OLM (https://github.com/operator-framework/operator-lifecycle-manager) : v0.19.0
+	// the total size checked is compressed
+	if bundle.CompressedSize > max_bundle_size {
+		errs = append(errs, errors.ErrInvalidBundle(
+			fmt.Sprintf("maximum bundle compressed size with gzip size exceeded: size=~%s , max=%s. Bundle uncompressed size is %s",
+				formatBytesInUnit(bundle.CompressedSize),
+				formatBytesInUnit(max_bundle_size),
+				formatBytesInUnit(bundle.Size)),
+			bundle.Name))
+	} else if float64(bundle.CompressedSize) > warnSize {
+		errs = append(errs, errors.WarnInvalidBundle(
+			fmt.Sprintf("nearing maximum bundle compressed size with gzip: size=~%s , max=%s. Bundle uncompressed size is %s",
+				formatBytesInUnit(bundle.CompressedSize),
+				formatBytesInUnit(max_bundle_size),
+				formatBytesInUnit(bundle.Size)),
+			bundle.Name))
 	}
 
 	return errs
+}
+
+func formatBytesInUnit(b int64) string {
+	const unit = 1000
+	if b < unit {
+		return fmt.Sprintf("%d B", b)
+	}
+	div, exp := int64(unit), 0
+	for n := b / unit; n >= unit; n /= unit {
+		div *= unit
+		exp++
+	}
+	return fmt.Sprintf("%.1f %cB",
+		float64(b)/float64(div), "kMGTPE"[exp])
 }
 
 // getBundleCRDKeys returns a set of definition keys for all CRDs in bundle.
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -48,4 +48,4 @@ jobs:
       with:
         fetch-depth: 0
     - name: Run go-apidiff
-      uses: joelanford/go-apidiff@master
+      uses: joelanford/go-apidiff@main
--- a/pkg/validation/internal/bundle_test.go
+++ b/pkg/validation/internal/bundle_test.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -164,7 +165,8 @@ func TestValidateServiceAccount(t *testing.T) {
 
 func TestBundleSize(t *testing.T) {
 	type args struct {
-		size int64
+		sizeCompressed int64
+		size           int64
 	}
 	tests := []struct {
 		name        string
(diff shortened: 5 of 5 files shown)


pkg/manifests/bundleloader.go
```diff
@@ -54,19 +54,32 @@ func (b *bundleLoader) calculateCompressedBundleSize() error {
 			if err != nil {
 				return err
 			}
+			if info.IsDir() {
+				return nil
+			}
+
+			if data, err := os.ReadFile(path); err == nil {
```
exdx, line 61: I'm not sure if the non-nil `err` from `os.ReadFile` is being returned anywhere? Maybe on L84 we should return the error (nil if reading the file went well) or the non-nil error instead of simply nil

pkg/validation/internal/bundle.go
```diff
@@ -137,14 +138,19 @@ func validateBundleSize(bundle *manifests.Bundle) []errors.Error {
...
+	}
+
+	if bundle.Size == nil || *bundle.Size == 0 {
+		errs = append(errs, errors.WarnFailedValidation("unable to check the bundle size", bundle.Name))
 		return errs
 	}
 
 	if *bundle.CompressedSize > max_bundle_size {
-		errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("maximum bundle compressed size with gzip size exceeded: size=~%d MegaByte, max=%d MegaByte", *bundle.CompressedSize/(1<<(10*2)), max_bundle_size/(1<<(10*2))), nil))
+		errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("maximum bundle compressed size with gzip size exceeded: size=~%d MegaByte, max=%d MegaByte. Bundle uncompressed size is %d KB", *bundle.CompressedSize/(1<<(10*2)), max_bundle_size/(1<<(10*2)), *bundle.Size/int64(1024)), bundle.Name))
```
exdx, line 151: maybe we could append the name of the bundle to the start of the error instead of the end so it's more clear, WDYT?
```suggestion
		errs = append(errs, errors.ErrInvalidBundle(fmt.Sprintf("maximum bundle compressed size with gzip size exceeded: bundle=%s, size=~%d MegaByte, max=%d MegaByte. Bundle uncompressed size is %d KB", bundle.Name, *bundle.CompressedSize/(1<<(10*2)), max_bundle_size/(1<<(10*2)), *bundle.Size/int64(1024))))
```

pkg/manifests/bundleloader.go
camilamacedo86: Done

pkg/validation/internal/bundle.go
camilamacedo86: We need to pass the name of the bundle otherwise it will be output as `Error : :` instead.

pkg/manifests/bundle.go
```diff
@@ -21,6 +21,8 @@ type Bundle struct {
 	Dependencies   []*Dependency
 	// CompressedSize stores the gzip size of the bundle
 	CompressedSize *int64
+	// Size stores the size of the bundle
```
bentito, line 24: nit: the indenting looks messed up for this struct chunk. Does running `go fmt` on it fix it maybe?

pkg/manifests/bundleloader.go
```diff
@@ -49,33 +49,58 @@ func (b *bundleLoader) LoadBundle() error {
...
+			if info.IsDir() {
+				return nil
+			}
+			data, err := os.ReadFile(path)
+			if err == nil {
+				fileSize, err := fileSize(path)
+				if err != nil {
+					return err
+				}
+				contentGzip, err := encoding.GzipBase64Encode(data)
```
bentito, line 69: Is the bundle compression really to Base64 or is it raw? The util method for gzip+base64 used here results in a human-readable, but potentially larger, compressed file. I see this allows us to check with `len` later which is convenient, but it's predicated on the base64 yielding text.

I will search for the code for the bundle compression and add a link here if I find it.

camilamacedo86: It is the same code. used by OLM
So, the idea here is we keep the same logic to check that.

pkg/manifests/bundle.go
camilamacedo86: I think it is solved. The ci checks the go fmt.

pkg/validation/internal/bundle.go
```diff
@@ -133,23 +134,66 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
 // - we could identify that the bundle size is close to the limit (bigger than 85%)
 func validateBundleSize(bundle *manifests.Bundle) []errors.Error {
 	warnPercent := 0.85
-	warnSize := int64(max_bundle_size * warnPercent)
+	warnSize := float64(max_bundle_size) * warnPercent
```
bentito, line 137: It's a little hard to tell here, but I think `warnSize` will come out as a `float64`. Later we compare it to an `CompressedSize` an `int64`. Does this yield reliable results?

camilamacedo86: We convert the compressedSize to float either see:`} else if float64(*bundle.CompressedSize) > warnSize {`

Also, we can check the tests so it shows fine: https://github.com/operator-framework/api/pull/229/files#diff-f4ded2e78dfeb583b32c94777fa213265c4748927288fb7c068d9e4b46e3fc5dR193-R197

pkg/manifests/bundleloader.go
bentito: I wasn't able to find in OLM repo searching on gzip...

pkg/validation/internal/bundle.go
bentito: ah, sorry, I read into the deleted code to see the unconverted compare.

pkg/validation/internal/bundle.go
```diff
@@ -133,23 +134,66 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
...
+	}
+	div, exp := int64(unit), 0
+	for n := b / unit; n >= unit; n /= unit {
+		div *= unit
+		exp++
+	}
+	return fmt.Sprintf("%.1f %cB",
+		float64(b)/float64(div), "kMGTPE"[exp])
+}
+
```
fgiloux, line 196: I see the intent but I am not too sure it is the best user experience. It possibly results in warnings / error messages that have different units for max, compressed and uncompressed values. I would favor having everything in KB.

pkg/validation/internal/bundle.go
```diff
@@ -133,23 +134,66 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
...
+	// Before these versions the bundles were not compressed
+	// and their size must be < ~1MB
+	if *bundle.Size > max_bundle_size {
+		errs = append(errs, errors.WarnInvalidBundle(
+			fmt.Sprintf("bundle uncompressed size exceeded the limit support for OLM versions relesed prior 1.17.5 :  size=~%s , max=%s. "+
+				"(these bundle cannot work in any cluster or vendor which uses OLM versions < 1.17.5 like OpenShift versions < 4.9)",
+				formatBytesInUnit(*bundle.Size),
+				formatBytesInUnit(max_bundle_size)),
+			bundle.Name))
```
fgiloux, line 177: I am not sure of the general approach and the code above being in the API repository.
Here is how I see things.
- Operator authors should have the tooling to validate the size of the bundles when they create / publish them: "operator-sdk validate" and "opm index add". There messages like above are very valuable.
- Once the bundle gets loaded onto an OpenShift cluster there are two options:
  - OLM < 1.17.5 no compression and the bundle size should be less than 1MB. But I guess the code above will never run on such a cluster or is there the intent to port it back but not the uncompression mechanism? Either ways the resulting uncompressed size is to be checked and that's it. I am not sure what value the above message brings to a cluster administrator.
  - OLM >= 1.17.5 the message above is rather noise for a cluster administrator

pkg/validation/internal/bundle.go
camilamacedo86: That is fine. I. am closing this one as sorted out.

pkg/validation/internal/bundle.go
camilamacedo86: Hi @fgiloux, 

The checks added here does not end up for cluster admins. They are consumed by the tools that have the goal to build and validate bundles only. 

The reason to add this WARNING is to better dealing/avoid use cases like: https://bugzilla.redhat.com/show_bug.cgi?id=2054514

Some advantages with: 

- Authors be aware that their bundle can work only if it will be published/distributed in OLM > = 1.17.5 (OCP > 4.9). That shows very important info for who publishes in OperatorHub.io
-  CI ( e.g. CVP reports): Help us easily identify why the deployment tests fail. The warning will appear in the report and reduce our effort in troubleshooting these scenarios. 
- For operator authors: help identify and avoid scenarios like: https://bugzilla.redhat.com/show_bug.cgi?id=2054514 or any scenario where they are not aware that their bundle cannot work on this versions which will be still in use for a while. 
- Give the ability for pipelines to use these warn if they be unable to test the bundle against all versions where it will be deployed. For e.g. we have pipelines that are testing only in the upper OLM version where the bundle will be distributed because of the high cost to test against a big matrix. 
- Audit: Give the ability to verify how many use cases we have published in the index and are not working at all because the bundle is > 1MB and these indexes are used with OLM versions <  1.17.5 (OCP < 4.9)

pkg/manifests/bundleloader.go
camilamacedo86: See; https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/encoding/encoding.go
It was moved to the API.
The next. step is to remove from OPM the implementation and import it from here.

pkg/validation/internal/bundle.go
fgiloux: @camilamacedo86 I know the background and understand the reason for the check. There are surely useful for tools that build and validate bundles. The thing is that this code as part of the API repository is imported by operator-lifecycle-manager, registry, etc and I fail to see why.
This repository is diverging more and more from what is written on the tin.

pkg/validation/internal/bundle.go
camilamacedo86: By testing/using it locally I checked that is how the SO usually does which makes it easier to verify/check the scenarios and compare.

pkg/validation/internal/bundle.go
camilamacedo86: It. is important for many projects because it is a lib. 
The validator is not used outside of this these operations 
Also, re-design the API shows out of scope. The goal of this PR is only to add the required fix and improvements to the check.

camilamacedo86: We spoke about this one and we agreed that we can close this one for now.Tks.

pkg/manifests/bundleloader.go
camilamacedo86: Marking as sorted out.

pkg/validation/internal/bundle.go
fgiloux: ack it is the way "ls -h" display things. No strong feeling about it. I feel like values are unlikely to diverge by a factor of thousand, hence my preference for having everything in KB but I am fine otherwise.

camilamacedo86: i will close this one as sorted out. But let's see if others came with similar thoughts as well.
We can push follow ups always

camilamacedo86:
/hold 
for @joelanford checks as well.

pkg/manifests/bundleloader.go
```diff
@@ -49,33 +49,58 @@ func (b *bundleLoader) LoadBundle() error {
...
 			}
+			if info.IsDir() {
+				return nil
+			}
+			data, err := os.ReadFile(path)
+			if err == nil {
+				fileSize, err := fileSize(path)
+				if err != nil {
+					return err
+				}
```
joelanford, line 68:
```suggestion
				fileSize := info.Size()
```

pkg/manifests/bundleloader.go
```diff
@@ -49,33 +49,58 @@ func (b *bundleLoader) LoadBundle() error {
...
 }
 
+func fileSize(path string) (int64, error) {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+
+	return fi.Size(), nil
+}
```
joelanford, line 102: This function can be dropped if we use `info.Size()` in the Walk func.

pkg/manifests/bundleloader.go
```diff
@@ -49,33 +49,58 @@ func (b *bundleLoader) LoadBundle() error {
...
-						return err
-					}
-					total := int64(len(content))
-					if b.bundle.CompressedSize != nil {
-						total += *b.bundle.CompressedSize
-					}
-					b.bundle.CompressedSize = &total
+				// Sum the bundle amount
+				totalSize := fileSize
+				if b.bundle.Size != nil {
```
joelanford, line 82: Move this above the compressed size calculation so its closer to where we read the uncompressed file size?

pkg/validation/internal/bundle.go
```diff
@@ -17,8 +17,9 @@ var BundleValidator interfaces.Validator = interfaces.ValidatorFunc(validateBund
 
 // max_bundle_size is the maximum size of a bundle in bytes.
 // This ensures the bundle can be staged in a single ConfigMap by OLM during installation.
-// The value is derived from the standard upper bound for k8s resources (~4MB).
-const max_bundle_size = 4 << (10 * 2)
+// The value is derived from the standard upper bound for k8s resources (~1MB).
+// We will use this value to check the bundle compressed is < ~1MB
+const max_bundle_size = int64(1 << (10 * 2))
```
joelanford, line 22: If a bundle is really close to this size, but still under the limit, it _could_ still fail because there is overhead in the structure of configmap.

Does it make sense to go look at what one of these configmaps looks like on cluster and estimate what a reasonable overhead number of bytes is so we can reduce this maximum size accordingly?

pkg/validation/internal/bundle.go
```diff
@@ -133,23 +134,66 @@ func getOwnedCustomResourceDefintionKeys(csv *operatorsv1alpha1.ClusterServiceVe
...
+		errs = append(errs, errors.WarnInvalidBundle(
+			fmt.Sprintf("nearing maximum bundle compressed size with gzip: size=~%s , max=%s. Bundle uncompressed size is %s",
+				formatBytesInUnit(*bundle.CompressedSize),
+				formatBytesInUnit(max_bundle_size),
+				formatBytesInUnit(*bundle.Size)),
+			bundle.Name))
+	}
+
+	// Before these versions the bundles were not compressed
+	// and their size must be < ~1MB
```
joelanford, line 170: I feel somewhat strongly that we shold keep the API repo focused on just one level of compatibility with OPM/OLM, so I'd suggest we don't check the uncompressed size at all since current versions of OPM/OLM do compression.

Otherwise, it seems like the API repo's main branch will start growing to encompass validation for all past history of what OPM/OLM support.

camilamacedo86: I will remove it and add in another PR so we are able to discuss it.
oo, we can move with the fix aasp.

pkg/manifests/bundleloader.go
```diff
@@ -49,26 +49,31 @@ func (b *bundleLoader) LoadBundle() error {
...
-						total += *b.bundle.CompressedSize
-					}
-					b.bundle.CompressedSize = &total
+				// Sum the compressed amount
+				contentGzip, err := encoding.GzipBase64Encode(data)
+				if err != nil {
+					return err
 				}
+				totalCompressed := int64(len(contentGzip))
+				b.bundle.CompressedSize += totalCompressed
```
joelanford, line 74: Nit: no need for `totalCompressed`, right?

pkg/validation/internal/bundle.go
camilamacedo86: After we begin to compress I think will be very hard we face this issue. 
Also, I am not sure if we would really have this overhead scenario. 

WDYT about we address here the fix and then, if we check this scenario or as a follow-up check if would be required consider it?

pkg/manifests/bundleloader.go
camilamacedo86: done

pkg/manifests/bundleloader.go
camilamacedo86: done

pkg/manifests/bundleloader.go
```diff
@@ -49,33 +49,58 @@ func (b *bundleLoader) LoadBundle() error {
...
+				contentGzip, err := encoding.GzipBase64Encode(data)
+				if err != nil {
+					return err
+				}
+				// Sum the compressed amount
+				totalCompressed := int64(len(contentGzip))
+				if b.bundle.CompressedSize != nil {
+					totalCompressed += *b.bundle.CompressedSize
+				}
+				b.bundle.CompressedSize = &totalCompressed
```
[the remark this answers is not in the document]
camilamacedo86: changed since we removed the pointer

pkg/validation/internal/bundle.go
joelanford: +1, this is an improvement we can make later.

pkg/manifests/bundleloader.go
bentito: :sgtm

PR #230, opened 2022-02-25
feat: add warn for bundles size < ~1MB

camilamacedo86:
**Description**

Add warning to let operator authors be aware that their bundle will not work on previous versions because are < ~1MB.

**Why?**

It has changed recently and authors who publish in OperatorHub as any other index can be aware that bundles > ~1MB does not work in the old versions. However, we can plan to remove this validation/warn in some timeframe like 1 yr. 

**Motivations**

- Authors be aware that their bundle can work only if it will be published/distributed in OLM > = 1.17.5 (OCP > 4.9). That shows very important info for who publishes in OperatorHub.io
-  CI ( e.g. CVP reports): Help us easily identify why the deployment tests fail. The warning will appear in the report and reduce our effort in troubleshooting these scenarios. 
- For operator authors: help identify and avoid scenarios like: https://bugzilla.redhat.com/show_bug.cgi?id=2054514 or any scenario where they are not aware that their bundle cannot work on this versions which will be still in use for a while. 
- Give the ability for pipelines to use these warn if they are unable to test the bundle against all versions where it will be deployed. For e.g. we have pipelines that are testing only in the upper OLM version where the bundle will be distributed because of the high cost to test against a big matrix. 
- Audit: Give the ability to verify how many use cases we have published in the index and are not working at all because the bundle is > 1MB and these indexes are used with OLM versions <  1.17.5 (OCP < 4.9)

### DIFF
--- a/pkg/validation/internal/bundle.go
+++ b/pkg/validation/internal/bundle.go
@@ -13,6 +13,25 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// BundleValidator implements Validator to validate Bundles.
+//
+// This check will verify if the Bundle spec is valid by checking:
+//
+// - for duplicate keys in the bundle, which may occur if a v1 and v1beta1 CRD of the same GVK appear.
+//
+// - if owned keys must matches with a CRD in bundle
+//
+// - if the bundle has APIs(CRDs) which are not defined in the CSV
+//
+// - if the bundle size compressed is < ~1MB
+//
+// NOTE: The bundle size test will raise an error if the size is bigger than the max allowed
+// and warnings when:
+// a) the api is unable to check the bundle size because we are running a check without load the bundle
+//
+// b) the api could identify that the bundle size is close to the limit (bigger than 85%)
+//
+// c) [Deprecated and planned to be removed at 2023 -  The API will start growing to encompass validation for all past history] - if the bundle size uncompressed < ~1MB and it cannot work on clusters which uses OLM versions < 1.17.5
 var BundleValidator interfaces.Validator = interfaces.ValidatorFunc(validateBundles)
 
 // max_bundle_size is the maximum size of a bundle in bytes.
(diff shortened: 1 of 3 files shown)


pkg/validation/internal/bundle.go
```diff
@@ -166,6 +166,18 @@ func validateBundleSize(bundle *manifests.Bundle) []errors.Error {
...
+	// todo: remove me at 2023
+	if bundle.Size > max_bundle_size {
+		errs = append(errs, errors.WarnInvalidBundle(
+			fmt.Sprintf("bundle uncompressed size exceeded the limit support for OLM versions relesed prior 1.17.5 :  size=~%s , max=%s. "+
+				"(these bundle cannot work in any cluster or vendor which uses OLM versions < 1.17.5 like OpenShift versions < 4.9)",
+				formatBytesInUnit(bundle.Size),
+				formatBytesInUnit(max_bundle_size)),
+			bundle.Name))
+	}
+
```
camilamacedo86, line 180: From @joelanford : 

> I feel somewhat strongly that we should keep the API repo focused on just one level of compatibility with OPM/OLM, so I'd suggest we don't check the uncompressed size at all since current versions of OPM/OLM do compression.

> Otherwise, it seems like the API repo's main branch will start growing to encompass validation for all past history of what OPM/OLM support.
I feel somewhat strongly that we should keep the API repo focused on just one level of compatibility with OPM/OLM, so I'd suggest we don't check the uncompressed size at all since current versions of OPM/OLM do compression.

Otherwise, it seems like the API repo's main branch will start growing to encompass validation for all past history of what OPM/OLM support.

camilamacedo86: I agree with your thought in general. But maybe is valid to think a little on this one because the change to compress which allows large bundles did not happen a long time ago and many authors publish their solution without being aware of this limitation. 

However, a strong con is we begin to provide expectations that we will begin to write code for old versions as you said. That is a really good reason for we drop it out.

camilamacedo86: /hold 

to think

camilamacedo86: @joelanford, 

WDYT? I tried to clarify the con raised by you and address this one as well. 
/hold cancel

joelanford: Isn't it kind of the same problem in other areas of Kubernetes? v1beta1 CRDs were removed in 1.22, so we went and removed all the code that supported v1beta1 CRDs in controller-runtime and controller-tools. If someone still needs to do something with v1beta1 CRDs we'd suggest using older versions of those tools.

IMO the same thing applies here. I'm guessing the friction is that we only use one version of the validator across all the versions of OLM that we support. Shouldn't we have a validator version per OLM version such that when we're validating a bundle for a particular version of OLM we use the validator version that goes with that OLM version?

camilamacedo86: We still parsing and allowing the v1beta1 in order to support old k8s versions.
However, fair enough. I really understand where you came from and I can go along with it.

PR #231, opened 2022-03-18
🐛 fix channel naming validation

camilamacedo86:
**Description**
- Move the warning to check the channel convention to the Good Practices validator which seems like to fit better
- Load/parse the annotations values from the yaml file to have the channels and default channel defined in the bundle 
- Add new test to ensure the scenarios

**Motivation**
Fix the issue:
https://github.com/operator-framework/operator-sdk/issues/5586
The warning is not working because we have not the values loaded from the bundle


camilamacedo86:
Hi @exdx,

Thank you for your help in the review. 
All done/addressed. please, feel free to check.

pkg/manifests/bundleloader.go
```diff
@@ -108,6 +118,19 @@ func (b *bundleLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err error)
 		return nil
 	}
 
+	annotationsFile := AnnotationsFile{}
+	if strings.HasPrefix(f.Name(), "annotations") {
+		annFile, err := readFile(path)
```
ryantking, line 123: Why not use https://pkg.go.dev/os#ReadFile ?

pkg/validation/internal/good_practices.go
```diff
@@ -75,3 +83,51 @@ func validateResourceRequests(csv *operatorsv1alpha1.ClusterServiceVersion) (err
...
+	for _, n := range array {
+		found := false
+		for _, v := range result {
+			if strings.TrimSpace(n) == strings.TrimSpace(v) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			result = append(result, n)
```
ryantking, line 128: Would `result = append(result, string.TrimSpace(n))` be better?

ryantking (commented):
Few minor potential changes

pkg/manifests/bundleloader.go
```diff
@@ -38,6 +39,15 @@ func (b *bundleLoader) LoadBundle() error {
 
 	errs = append(errs, b.calculateCompressedBundleSize())
 
+	// Set values from the annotations when the values are not loaded
```
awgreene, line 42: This seems fine, but do we document that channels/default channels set in the bundle are prioritized over those set in the bundle?

camilamacedo86: Currently, it has not been populated in any way. The default and channels are never populated and that was the issue that we are fixing here. We actually do not need the do the check conditional here.

pkg/validation/internal/good_practices.go
awgreene: You could also use a `map[string]struct{}` to track names you've seen.

pkg/validation/internal/good_practices.go
```diff
@@ -75,3 +83,51 @@ func validateResourceRequests(csv *operatorsv1alpha1.ClusterServiceVersion) (err
...
+			!strings.HasPrefix(channel, stable) &&
+			!strings.HasPrefix(channel, fast) &&
+			channel != "" {
+			channelsNotFollowingConventional = append(channelsNotFollowingConventional, channel)
+		}
+
+	}
+
+	if len(channelsNotFollowingConventional) > 0 {
+		return fmt.Errorf("channel(s) %+q are not following the recommended naming convention: "+
```
awgreene, line 108: Should this point to https://olm.operatorframework.io/docs/best-practices/channel-naming/#recommended-channel-naming instead?

pkg/manifests/bundleloader.go
camilamacedo86: Good catcher :-)

pkg/manifests/bundleloader.go
camilamacedo86: Added a comment to clarify it in the code.

pkg/validation/internal/operatorhub.go
```diff
@@ -215,40 +215,9 @@ func validateBundleOperatorHub(bundle *manifests.Bundle, k8sVersion string) erro
...
-		result.Add(errors.WarnFailedValidation(warn.Error(), bundle.CSV.GetName()))
-	}
-
 	return result
 }
 
-// validateHubChannels will check the channels. The motivation for the following check is to ensure that operators
-// authors knows if their operator bundles are or not respecting the Naming Convention Rules.
-// However, the operator authors still able to choose the names as please them.
-func validateHubChannels(channels []string) error {
```
awgreene, line 228: Just to be clear, we're moving this to the "good practices" file since these are suggested best channel naming conventions, right?

pkg/validation/internal/testdata/bundle_with_metadata/manifests/cache.example.com_memcacheds.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apiextensions.k8s.io/v1
```
awgreene, line 1: Non-blocking nit: I wish we had a slimmer operator for testing purposes.

pkg/validation/internal/good_practices.go
camilamacedo86: The whole doc is about this topic.
So, I would prefer to link the doc instead of a topic from the doc since is more unlike we change the doc link then remove/rename some section on it.

pkg/validation/internal/operatorhub.go
camilamacedo86: Yes. It was added to the OperatorHub because we had not the good practices one.
However, it shows fit better in this Validator since we might want to call only this one to check all recommendations related to common suggestions and good practices.

ryantking (commented):
Small change then good to approve

pkg/validation/internal/good_practices.go
```diff
@@ -75,3 +83,46 @@ func validateResourceRequests(csv *operatorsv1alpha1.ClusterServiceVersion) (err
...
+	var result []string
+	uniqueValues := make(map[string]string)
+	for _, n := range array {
+		uniqueValues[strings.TrimSpace(n)] = ""
+	}
+
+	for k, _ := range uniqueValues {
+		result = append(result, k)
+	}
+	return result
```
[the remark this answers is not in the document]
camilamacedo86: that is a nit that shows outside of the fixed scope.
Can we agree on doing that in a follow-up?

ryantking (approved):

dinhxuanvu (approved):
/lgtm

PR #232, opened 2022-03-18
✨ change: move check bundle name to the Good Practices validator

camilamacedo86:
**Description**
change: move check bundle name to the Good Practices validator

**Motivation**
Now, that we have the Good Practices one the warning when the bundle name when is not following the semantics `<operator-name>.v<semver>` shows fit better in this category of checks.


pkg/validation/internal/good_practices.go
```diff
@@ -3,9 +3,11 @@ package internal
 import (
 	goerrors "errors"
 	"fmt"
+	"github.com/blang/semver/v4"
```
exdx, line 6: nit: imports

pkg/validation/internal/good_practices_test.go
```diff
@@ -92,3 +93,84 @@ func Test_ValidateGoodPractices(t *testing.T) {
...
+	tests := []struct {
+		name        string
+		args        args
+		wantError   bool
+		wantWarning bool
+		errStrings  []string
+		warnStrings []string
+	}{
+		{
+			name: "should work with valid bundle name",
```
exdx, line 110: could we add a test case for patch releases? ie `memcached-operator.v0.9.2+alpha` just to see if it works?

pkg/validation/internal/good_practices.go
```diff
@@ -69,9 +78,37 @@ func validateResourceRequests(csv *operatorsv1alpha1.ClusterServiceVersion) (err
...
+			checks.warns = append(checks.warns, fmt.Errorf("csv.metadata.Name %v is not following the versioning "+
+				"convention (MAJOR.MINOR.PATCH e.g 0.0.1): https://semver.org/", checks.csv.Name))
+		}
+	} else {
+		checks.warns = append(checks.warns, fmt.Errorf("csv.metadata.Name %v is not following the versioning "+
+			"convention (MAJOR.MINOR.PATCH e.g 0.0.1): https://semver.org/", checks.csv.Name))
+	}
+
+	// Check if its following the name convention
+	if len(strings.Split(checks.csv.Name, ".v")) < 2 {
```
exdx, line 108: what if the length is greater than two? like `memcached-operator.v0.0.1.vfoo`, this warning wouldn't occur -- do we want to have the check where the `len != 2` instead -- is that safer?

PR #233, opened 2022-03-18
✨ feat: new validator to check Multiple Architectures configuration

camilamacedo86:
**Description**
add: new validator to check Multiarch configuration

**Motivation**
Help users, cis and tools verify the  Multiarch configuration

### DIFF
--- a/pkg/validation/internal/multiarch.go
+++ b/pkg/validation/internal/multiarch.go
@@ -70,7 +70,7 @@ type multiArchValidator struct {
 	managerImagesString []string
 	// managerArchs contains a map of the arch types found
 	managerArchs map[string]string
-	// managerOs contains a map of the so(s) found
+	// managerOs contains a map of the OSes found
 	managerOs map[string]string
 	// Store the bundle load
 	bundle *manifests.Bundle
--- a/pkg/validation/internal/multiarch.go
+++ b/pkg/validation/internal/multiarch.go
@@ -1,0 +1,623 @@
+package internal
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"sort"
+	"strings"
+
+	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
+// MultipleArchitecturesValidator validates the bundle against criteria to support Multiple Architectures. For further
+// information check: https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/
+//
+// This validator will inspect the images with the chosen container-tool. One of: [docker, podman, none] (By default docker)
+// and then: (It is only used to $container-tool manifest inspect)
+//
+// - raise a error(s) when is possible to confirm that images do not provide the support defined via to the labels in the CSV
+//
+// - raise a warning when it is possible to check that the Operator manager image(s) supports architecture(s) not defined via labels. Therefore, it shows like the labels are missing.
+//
+// - raise warnings when it is possible to verify that the images defined in the CSV do not provide the same architecture(s) supported by the Operator manager image(s) or defined via the labels
+//
+// ### What is checked?
+//
+// On this check, we aggregate the archetype(s) and OS(s) provided via the labels and those which are found by checking the images so that, we can check:
+//
+// - If your CSV is missing labels
+//
+// - If your Operator bundle specifies images which do not support all archetypes found for your Operator image(s) (probably supported by your project)
+//
+// Note: To better guess the case scenarios where authors might have missed the labels, the following check will verify all architectures supported by the Operator image(s). However, by looking at the CSV we are not able to ensure what is the Operator image because this info is not provided. Therefore, we know by SDK the Operator image container will be called manager.
+//
+// ### How the Operator image(s) are identified?
+//
+// The container named as manager under the CSV Deployment InstallStrategy (`Spec.InstallStrategy.StrategySpec.DeploymentSpecs`)
+// And if the above not found, all images under the InstallStrategy excluding the container named as `kube-rbac-proxy` since it is also scaffolded by default via SDK
+var MultipleArchitecturesValidator interfaces.Validator = interfaces.ValidatorFunc(multipleArchitecturesValidate)
+
+// ContainerToolsKey defines the key which can be used by its consumers
+// to inform where to find the container tool that should be used to inspect the image
+const ContainerToolsKey = "container-tools"
+
+// operatorFrameworkArchLabel defines the label used to store the supported Arch on CSV
+const operatorFrameworkArchLabel = "operatorframework.io/arch."
+
+// operatorFrameworkOSLabel stores the labels for the supported OS from the CSV
+const operatorFrameworkOSLabel = "operatorframework.io/os."
+
+// default_container_scaffold_by_sdk defines the name of a default scaffold done by SDK
+// it is useful for we are able to find the operator manager image more assertively
+const default_container_scaffold_by_sdk = "kube-rbac-proxy"
+
+// multiArchValidator store the data to perform the tests
+type multiArchValidator struct {
+	// infraCSVArchLabels store the arch labels (i.e amd64, ppc64le) from
+	// operatorframework.io/arch.<GOARCH>: supported
+	infraCSVArchLabels []string
+	// InfraCVSOSLabels store the OS labels from
+	// operatorframework.io/os.<GOARCH>: supported
+	infraCSVOSLabels []string
+	// allOtherImages stores the allOtherImages defined in the bundle with the platform.arch supported
+	allOtherImages map[string][]platform
+	// managerImages stores the images that we could consider as from the manager
+	managerImages map[string][]platform
+	// managerImagesString stores the images only
+	managerImagesString []string
+	// managerArchs contains a map of the arch types found
+	managerArchs map[string]string
+	// managerOs contains a map of the so(s) found
+	managerOs map[string]string
+	// Store the bundle load
+	bundle *manifests.Bundle
+	// containerTool defines the container tool which will be used to inspect the allOtherImages
+	containerTool string
+	// warns stores the errors faced by the validator to return the warnings
+	warns []error
+	// warns stores the errors faced by the validator to return the warnings
+	errors []error
+}
+
+// manifestInspect store the data obtained by running container-tool manifest inspect <IMAGE>
+type manifestInspect struct {
+	ManifestData []manifestData `json:"manifests"`
+}
+
+// manifestData store the platforms
+type manifestData struct {
+	Platform platform `json:"platform"`
+}
+
+// platform store the Architecture and OS supported by the image
+type platform struct {
+	Architecture string `json:"architecture"`
+	OS           string `json:"os"`
+}
+
+func multipleArchitecturesValidate(objs ...interface{}) (results []errors.ManifestResult) {
+	// Obtain the k8s version if informed via the objects an optional
+	var containerTool = ""
+	for _, obj := range objs {
+		switch obj.(type) {
+		case map[string]string:
+			// Check the key values informed
+			containerTool = obj.(map[string]string)[ContainerToolsKey]
+			if len(containerTool) > 0 {
+				// Make lower for we compare and use it
+				containerTool = strings.ToLower(containerTool)
+				break
+			}
+		}
+	}
+
+	for _, obj := range objs {
+		switch v := obj.(type) {
+		case *manifests.Bundle:
+			results = append(results, validateMultiArchWith(v, containerTool))
+		}
+	}
+	return results
+}
+
+func validateMultiArchWith(bundle *manifests.Bundle, containerTool string) errors.ManifestResult {
+	result := errors.ManifestResult{}
+	if bundle == nil {
+		result.Add(errors.ErrInvalidBundle("bundle is nil", nil))
+		return result
+	}
+
+	result.Name = bundle.Name
+
+	if bundle.CSV == nil {
+		result.Add(errors.ErrInvalidBundle("bundle csv is nil", bundle.Name))
+		return result
+	}
+
+	// Validate inputs. If a container-tool key be informed
+	// with an invalid/unsupported value then make no sense do the check
+	containerTool, err := validateContainerTool(containerTool)
+	if err != nil {
+		result.Add(errors.ErrFailedValidation(err.Error(), bundle.CSV.GetName()))
+		return result
+	}
+
+	// Performs the checks
+
+	multiArchValidator := multiArchValidator{bundle: bundle, containerTool: containerTool}
+	multiArchValidator.validate()
+
+	for _, err := range multiArchValidator.warns {
+		// add the warn to the result
+		result.Add(errors.WarnFailedValidation(err.Error(), bundle.CSV.GetName()))
+	}
+
+	for _, err := range multiArchValidator.errors {
+		// add the warn to the result
+		result.Add(errors.ErrFailedValidation(err.Error(), bundle.CSV.GetName()))
+	}
+
+	return result
+}
+
+// validateContainerTool verifies if the container tool informed is valid
+func validateContainerTool(containerTool string) (string, error) {
+	if len(containerTool) == 0 || containerTool == "none" {
+		containerTool = "docker"
+	} else if containerTool != "docker" && containerTool != "podman" {
+		return containerTool, fmt.Errorf("invalid value (%s) for (%s). One of: [docker, podman, none] "+
+			"(If not set, the default value is docker)", ContainerToolsKey, containerTool)
+	}
+	return containerTool, nil
+}
+
+// validate performs all required checks to validate the bundle against the Multiple Architecture
+// configuration to guess the missing labels and/or highlight what are the missing Architectures
+// for the allOtherImages (for what is configured to be supported AND for what we guess that is supported
+// and just is missing a label).
+func (data *multiArchValidator) validate() {
+	data.loadInfraLabelsFromCSV()
+	data.loadImagesFromCSV()
+	data.managerImages = data.inspectImages(data.managerImages)
+	data.allOtherImages = data.inspectImages(data.allOtherImages)
+	data.loadAllPossibleArchSupported()
+	data.loadAllPossibleSoSupported()
+	data.doChecks()
+}
+
+// loadInfraLabelsFromCSV will gather the respective labels from the CSV
+func (data *multiArchValidator) loadInfraLabelsFromCSV() {
+	data.managerArchs = make(map[string]string)
+	data.managerOs = make(map[string]string)
+
+	for k, v := range data.bundle.CSV.ObjectMeta.Labels {
+		if strings.Contains(k, operatorFrameworkArchLabel) && v == "supported" {
+			data.infraCSVArchLabels = append(data.infraCSVArchLabels, k)
+		}
+	}
+	for k, v := range data.bundle.CSV.ObjectMeta.Labels {
+		if strings.Contains(k, operatorFrameworkOSLabel) && v == "supported" {
+			data.infraCSVOSLabels = append(data.infraCSVOSLabels, k)
+		}
+	}
+}
+
+// loadImagesFromCSV will add all allOtherImages found in the CSV
+// it will be looking for all containers allOtherImages and what is defined
+// via the spec.relatedImages (required for disconnect support)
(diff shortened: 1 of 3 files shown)


pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+	// Obtain the k8s version if informed via the objects an optional
+	var containerTool = ""
+	for _, obj := range objs {
+		switch obj.(type) {
+		case map[string]string:
+			containerTool = obj.(map[string]string)[ContainerToolsKey]
+			if len(containerTool) > 0 {
+				break
+			}
+		}
```
ryantking, line 95: I think the following is more clear:

```go
if objMap, ok := obj.(map[string]string); ok && containerTool == “” {
    containerTool = objMap[ContainerToolsKey]
    break
}
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	for _, obj := range objs {
+		switch v := obj.(type) {
+		case *manifests.Bundle:
+			results = append(results, validateMultiArchWith(v, containerTool))
+		}
```
ryantking, line 102: Similar comment to above:
```go
if bundle, ok := obj.(*manifests.Bundle); ok {
    // append bundle to results
}
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+	if bundle.CSV == nil {
+		result.Add(errors.ErrInvalidBundle("Bundle csv is nil", bundle.Name))
+		return result
+	}
+
+	containerTool, err := validateContainerTool(containerTool)
+	if err != nil {
+		result.Add(errors.ErrFailedValidation(err.Error(), bundle.CSV.GetName()))
+		return result
+	}
```
ryantking, line 125: I think we should move this block up to `multipleArchitecturesValidate`. This function does not actually use `containerTool`, it just accepts it as a parameter and passes it into the validator. In my view, the function that retrieves the `containerTool` value should be the one to validate it and return an error.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+
+	multiArchValidator := MultiArchValidator{Bundle: bundle, ContainerTool: containerTool}
+	multiArchValidator.validate()
+
+	for _, err := range multiArchValidator.Warns {
+		result.Add(errors.WarnFailedValidation(err.Error(), bundle.CSV.GetName()))
+	}
```
ryantking, line 132: I think this section is a little bit confusing.

1. Why is `MultiArchValidator` exported when its primary function, `validate` is not?
2. Why are `Warns` read off the struct instead of being returned?

Could we do something like this?

```go
for _, err := range ValidateMultiArchBundle(bundle, containerTool) {
    // add the warn to the result
}

func ValidateMultiArchBundle(bundle *manifest.Bundle, containerTool string) []error {
    data := &multiArchValicatorData{
        AllArchFound: make(map[string]string),
        // etc
    }
    data.getInfraLabelsFromCSV(bundle)
    return data.checks() // this will return the warnings instead of mutating the struct
}
```

This gives us:

1. Single entrypoint to validate function, no instantiating struct then calling validate.

2. More self documenting since consumers of the API won’t need to know to read struct fields to get results. 

3. Don’t need to use single use struct.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+
+	return result
+}
+
+// validateContainerTool verifies if the container tool informed is valid
+func validateContainerTool(containerTool string) (string, error) {
+	if len(containerTool) == 0 || containerTool == "none" {
```
ryantking, line 139:
```suggestion
	if len(containerTool) == 0 || strings.ToLower(containerTool) == "none" {
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+		data.Images[v.Image] = append(data.Images[v.Image], Platform{})
+	}
+}
+
+// runManifestInspect executes the command for we are able to check what
+// are the Architecture(s) and SO(s) supported per each image found
+func runManifestInspect(image, tool string) (ManifestInspect, error) {
+	cmd := exec.Command(tool, "manifest", "inspect", image)
+	output, err := runCommand(cmd)
+	if err != nil || len(output) < 1 {
```
ryantking, line 200: In the case that `err` is `nil` and the length of the output is 0, you’ll get a weird silent failure. I would prefer:

```go
if err != nil {
    // return err
} else if len(output) == 0 {
    // return empty output error
}
```

The current code will return an empty manifest and nil error when the output is empty.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+func (data *MultiArchValidator) checkMissingImagesSupport() {
+	for image, arc := range data.Images {
+		listArchNotFound := []string{}
+		for archFromList := range data.AllArchFound {
+			found := false
+			for _, imageData := range arc {
+				// Ignore the case when the Plataform.Architecture == "error" since that means
+				// that was not possible to inspect the image
+				if imageData.Architecture == "error" {
+					found = true
```
ryantking, line 266: What do we gain by not breaking here? The only thing that can happen is `found` gets set to `true` a second time.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+				break
+			}
+		}
+		if !found {
+			// If the value is linux and no labels were added to the CSV then it is fine
+			if !(supported == "linux" && len(data.InfraSOLabels) == 0) {
+				notFoundSoLabel = append(notFoundSoLabel, supported)
+			}
+
+		}
```
ryantking, line 337:
```suggestion
		// If the value is linux and no labels were added to the CSV then it is fine
		if !found && !(supported == "linux" && len(data.InfraSOLabels) == 0) {
			notFoundSoLabel = append(notFoundSoLabel, supported)
		}
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,513 @@
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			// If the value is amd64 and no labels were added to the CSV then it is fine
+			if !(supported == "amd64" && len(data.InfraArchLabels) == 1) {
+				notFoundArchLabel = append(notFoundArchLabel, supported)
+			}
+		}
```
ryantking, line 372: Similar suggestion to above

ryantking (changes requested):
Few changes and questions. A lot of it is stylistic stuff to make the code more readable.

pkg/validation/internal/multiarch.go
camilamacedo86: The idea is valid for all input first. We could have other keys. 
So,  would not make sense to move forward without first checking if the input is correct. 
See that if we do what you suggest we will start to do the checks before ensuring that it is actually possible with the data provided.

pkg/validation/internal/multiarch.go
camilamacedo86: Not sure if that would be more clear, what happens if we have more optional keys?

pkg/validation/internal/multiarch.go
camilamacedo86: same here. If we have other types? 
I think I would prefer to stick with the current implementation which indeed is == across all validators.

pkg/validation/internal/multiarch.go
camilamacedo86: done some improvements based on your feedback.

pkg/validation/internal/multiarch.go
camilamacedo86: Good point. 
Changed on top for we ensure that we will have the lower case always.

pkg/validation/internal/multiarch.go
camilamacedo86: For both, the return is always the same `return manifestInspect{}, err`.
Not sure if I could follow up on your suggestion.

pkg/validation/internal/multiarch.go
camilamacedo86: So, I tried to clarify in the code and I am closing this one as solved let me know if you think that we should re-open.

pkg/validation/internal/multiarch.go
ryantking: The situation I'm thinking of is when we have:
```go
output := ""
err := nil
```
The return value of `runManifestInspect` will be `ManifestInspect{}, nil`, which will be confusing since the caller will need to know the meaning of the `ManifestInspect` zero value. I think we would want `ManifestInspect{}, errors.New("no output returned from manifest tool")` to make it clear what happened.

pkg/validation/internal/multiarch.go
camilamacedo86: It is totally possible.

pkg/validation/internal/multiarch.go
camilamacedo86: The `no output returned from manifest tool` would return an error always in this case. 
How can we do a docker manifest inspect and the manifest be parsed and empty and no error returned?

ryantking: Then why check for the empty output in the first place? From what you just said if `len(output) == 0` is `true` then `err != nil` will also be `true`.

I also don't want to hold up this PR on this relatively small edge case, so I am good to approve if you are confident that we need to check the output length and it will never be empty without an error, but I stand by that you should handle `len(output) == 0` in a separate `if/else` branch or drop the check entirely.

camilamacedo86: I think you are right the len(output) == 0 is redundant.
Removed :-)

camilamacedo86: I think you are right the len(output) == 0 is redundant.
Removed :-)

camilamacedo86: I think you are right the len(output) == 0 is redundant.
Removed :-)

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,521 @@
...
+				containerTool = strings.ToLower(containerTool)
+				break
+			}
+		}
+	}
+
+	for _, obj := range objs {
+		switch v := obj.(type) {
+		case *manifests.Bundle:
+			results = append(results, validateMultiArchWith(v, containerTool))
```
jmrodri, line 104: Why are we looping over the objs twice? Can we not do both checks in a single iteration?

jmrodri (changes requested):
A couple suggestions and question about SO vs OS.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,521 @@
...
+const operatorFrameworkSoLabel = "operatorframework.io/so."
+
+// multiArchValidator store the data to perform the tests
+type multiArchValidator struct {
+	// infraArchLabels store the arch labels (i.e amd64, ppc64le) from
+	// operatorframework.io/arch.<GOARCH>: supported
+	infraArchLabels []string
+	// InfraSOLabel store the OS labels from
+	// operatorframework.io/so.<GOARCH>: supported
+	infraSOLabels []string
```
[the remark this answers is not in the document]
camilamacedo86: Really good catcher 👍 🥇 THANK YOU. 
All should be os + the label is "operatorframework.io/os"

pkg/validation/internal/multiarch.go
camilamacedo86: If we put both together then can we not call the validate before getting the key?
So that is done first to get the key after we call the validate
Note that all are == across the board and I agree that we have a great hall for improvements.
However, wdyt about we let this for when we change one place we do == for all the other places?

pkg/validation/internal/multiarch.go
camilamacedo86: See:

```
// Check the key values informed
			containerTool = obj.(map[string]string)[ContainerToolsKey]
			if len(containerTool) > 0 {
				// Make lower for we compare and use it
				containerTool = strings.ToLower(containerTool)
				break
			}
```
We ensure that the string is informed as the value of the key will be the lower case always so that here we just check if the value is or is not valid. Is that make sense?

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,521 @@
...
+					"Then, it is very likely that you want to provide it",
+					image, listAllSoNotFound))
+		}
+	}
+
+}
+
+// verify if 1 or ore images has support for a SO not defined via the labels
+// (probably the label for this SO is missing )
+func (data *multiArchValidator) checkMissingLabelsForSO() {
```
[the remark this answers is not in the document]
camilamacedo86: should all be so right? :-) Typo issue

camilamacedo86:
@jmrodri THANK YOU a lot.
All addressed, please feel free to review it again.

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,532 @@
...
+	data.loadImagesFromCSV()
+	data.inspectAllImages()
+	data.loadAllPossibleArchSupported()
+	data.loadAllPossibleSoSupported()
+	data.doChecks()
+}
+
+// loadInfraLabelsFromCSV will gather the respective labels from the CSV
+func (data *multiArchValidator) loadInfraLabelsFromCSV() {
+	for k, v := range data.bundle.CSV.ObjectMeta.Labels {
```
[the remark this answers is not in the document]
camilamacedo86: That was a good catcher. But the label is `Labels map[string]string`.
So that do not panic if be nil. See: https://go.dev/play/p/VUZPduMAOIf

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,532 @@
...
+		if strings.Contains(k, operatorFrameworkOSLabel) && v == "supported" {
+			data.infraOSLabels = append(data.infraOSLabels, k)
+		}
+	}
+}
+
+// loadImagesFromCSV will add all images found in the CSV
+// it will be looking for all containers images and what is defined
+// via the spec.RELATE_IMAGE (required for disconnect support)
```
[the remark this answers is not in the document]
camilamacedo86: since that I small nit if we have not other ones and if you do not mind I prefer to address in a follow up for we do lost the lgtm label :-)

ryantking (approved):

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+
+	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
+// MultipleArchitecturesValidator validates the bundle against criteria to support Multiple Architectures. For further
+// information check: https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/
+//
+// This validator will inspect the images with the container-tool informed. One of: [docker, podman, none] (By default docker)
```
perdasilva, line 18:
```suggestion
// This validator will inspect the images with the chosen container-tool. One of: [docker, podman, none] (By default docker)
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
+// MultipleArchitecturesValidator validates the bundle against criteria to support Multiple Architectures. For further
+// information check: https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/
+//
+// This validator will inspect the images with the chosen container-tool. One of: [docker, podman, none] (By default docker)
+// and then: (It is only used to $container-tool manifest inspect)
+//
+// - raise a error(s) when is possible to confirm that images does not provide the support defined via to the labels on the CSV
```
bentito, line 21:
```suggestion
// - raise a error(s) when is possible to confirm that images do not provide the support defined via to the labels on the CSV
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+//
+// - raise warnings when is possible to verify that the images defined in the CSV does not provide the same architecture(s) supported by the Operator manager image(s) or defined via the labels
+//
+// ### What is checked?
+//
+// On this check, we aggregate the archetype(s) and OS(s) provided via the labels and those which are found by checking the images so that, we can check:
+//
+// - If your CSV is missing labels
+//
+// - If your Operator bundle specifies images which does not supports all archetypes found for your Operator image(s) (probably supported by your project)
```
bentito, line 33:
```suggestion
// - If your Operator bundle specifies images which do not supports all archetypes found for your Operator image(s) (probably supported by your project)
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+//
+// Note: To better guess the case scenarios where authors might missed the labels the following check will verify all architectures support for the Operator image(s). However, by looking at the CSV we are not able to ensure what is the Operator image because this info is not provided. Therefore, we know by SDK the Operator image container will be called manager.
```
bentito, line 35:
```suggestion
// Note: To better guess the case scenarios where authors might have missed the labels, the following check will verify all architectures supported by the Operator image(s). However, by looking at the CSV we are not able to ensure what is the Operator image because this info is not provided. Therefore, we know by SDK the Operator image container will be called manager.
```

bentito: I'm still not sure what is meant here?

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+var MultipleArchitecturesValidator interfaces.Validator = interfaces.ValidatorFunc(multipleArchitecturesValidate)
+
+// ContainerToolsKey defines the key which can be used by its consumers
+// to inform where their container tool that should be used to inspect the image
+const ContainerToolsKey = "container-tools"
+
+// operatorFrameworkArchLabel defines the label used to store the supported Arch on CSV
+const operatorFrameworkArchLabel = "operatorframework.io/arch."
+
+// operatorFrameworkOSLabel defines the label used to store the supported So on CSV
```
bentito, line 50:
```suggestion
// operatorFrameworkOSLabel stores the labels for the supported OS from the CSV
```

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+//
+// ### How the Operator image(s) are identified?
+//
+// The container named as manager under the CSV Deployment InstallStrategy (`Spec.InstallStrategy.StrategySpec.DeploymentSpecs`)
+// And if the above not found, all images under the InstallStrategy excluding the container named as `kube-rbac-proxy` since it is also scaffolded by default via SDK
+var MultipleArchitecturesValidator interfaces.Validator = interfaces.ValidatorFunc(multipleArchitecturesValidate)
+
+// ContainerToolsKey defines the key which can be used by its consumers
+// to inform where their container tool that should be used to inspect the image
+const ContainerToolsKey = "container-tools"
```
bentito, line 45: Would `ContainerToolsPath` be a better name?

camilamacedo86: I do not think so because it is a key, value map.
We have the same in other places so I would vote in keep all following the same standards.

bentito: ok

pkg/validation/internal/multiarch.go
```diff
@@ -0,0 +1,623 @@
...
+		// add the warn to the result
+		result.Add(errors.ErrFailedValidation(err.Error(), bundle.CSV.GetName()))
+	}
+
+	return result
+}
+
+// validateContainerTool verifies if the container tool informed is valid
+func validateContainerTool(containerTool string) (string, error) {
+	if len(containerTool) == 0 || containerTool == "none" {
```
bentito, line 168: But isn't "none" explicitly saying: "there is no container tool for this"? If that's the case, it should not ignore that setting and default to Docker.

camilamacedo86: This is behaviour across the tools. (Keeping the same) None is == "empty"
Could you please check the commands in OPM and SDK that allows us to pass the container tools (i.e.: https://github.com/operator-framework/operator-sdk/blob/5f6c16c792b486c5d955a36b060112be288443ff/website/content/en/docs/olm-integration/cli-overview.md#private-bundle-and-catalog-image-registries)

bentito: ah okay

dinhxuanvu (approved):
/approve

PR #234, opened 2022-04-04
Add CRD description check validation

ryantking:
This adds a check to the good practices validator that ensures all CRDs
declared in the bundle have non-empty descriptions. Both owned and
required CRDs will be checked. Note that this validates the CRD
definition in the CSV, not any field in the CRD itself.

Signed-off-by: Ryan King <ryking@redhat.com>


pkg/validation/internal/good_practices.go
```diff
@@ -54,6 +54,9 @@ func validateGoodPracticesFrom(bundle *manifests.Bundle) errors.ManifestResult {
 	for _, warn := range warns {
 		result.Add(errors.WarnFailedValidation(warn.Error(), bundle.CSV.GetName()))
 	}
+	for _, err := range validateCrdDescriptions(bundle.CSV.Spec.CustomResourceDefinitions) {
+		result.Add(errors.ErrFailedValidation(err.Error(), bundle.CSV.GetName()))
```
camilamacedo86, line 58: We need return a Warning in this scenario.
```suggestion
		result.Add(errors.WarnFailedValidation(err.Error(), bundle.CSV.GetName()))
```
We cannot fail with an error because that means blocking new distributions when we should not.

pkg/validation/internal/good_practices.go
```diff
@@ -126,3 +129,18 @@ func getUniqueValues(array []string) []string {
...
+		for _, crd := range crds {
+			if crd.Description == "" {
+				errors = append(errors, fmt.Errorf("%s CRD %q has an empty description", relation, crd.Name))
+			}
+		}
+		return errors
+	}
+
+	return append(f(crds.Owned, "owned"), f(crds.Required, "required")...)
+}
```
camilamacedo86, line 146: We also need to add a test for this one in: https://github.com/operator-framework/api/blob/master/pkg/validation/internal/good_practices_test.go

camilamacedo86: See, e.g; https://github.com/operator-framework/api/blob/master/pkg/validation/internal/good_practices_test.go

camilamacedo86 (approved):
That shows great for me 👍

pkg/validation/internal/good_practices.go
camilamacedo86: feel free to close sorted out.

pkg/validation/internal/good_practices.go
camilamacedo86: feel free to close sorted out. :-)

jmrodri (approved):
/lgtm

PR #235, opened 2022-04-04
Introduce UpgradeStrategy to OperatorGroups

awgreene:


awgreene:
/hold until @joelanford can review.

An example of the kubectl explain command can be found below:
```bash
$ kubectl explain operatorgroup.spec.upgradeStrategy
KIND:     OperatorGroup
VERSION:  operators.coreos.com/v1

RESOURCE: upgradeStrategy <Object>

DESCRIPTION:
     UpgradeStrategy defines the upgrade strategy for operators in the
     namespace. There are currently two supported upgrade strategies:
     Default: OLM will only allow ClusterServiceVersions to move to the
     replacing phase from the succeeded phase. This effectively means that OLM
     will not allow operators to move to the next version if an installation or
     upgrade has failed. UnsafeFailForward: OLM will allow
     ClusterServiceVersions to move to the replacing phase from the succeeded
     phase or from the failed phase. Additionally, OLM will generate new
     instalplans when a subscription references a failed InstallPlan and the
     catalog has been updated with a new upgrade for the existing set of
     operators. This upgrade strategy is unsafe and may result in dataloss
     unless you have total control and deep understanding of the set of
     operators being managed in the namespace. The OLM team does not recommend
     using this feature in production environments.

FIELDS:
   name	<string>
```

grokspawn:
Kill the empty tmp.txt? 
Otherwise /lgtm

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -45,6 +45,49 @@ type OperatorGroupSpec struct {
 	// Static tells OLM not to update the OperatorGroup's providedAPIs annotation
 	// +optional
 	StaticProvidedAPIs bool `json:"staticProvidedAPIs,omitempty"`
+
+	UpgradeStrategy *UpgradeStrategy `json:"upgradeStrategy,omitempty"`
```
joelanford, line 49: Since we have a default value for this setup in the CRD schema, WDYT about making this field always present, such that we can always apply the default `name: Default`?

```suggestion
	UpgradeStrategy UpgradeStrategy `json:"upgradeStrategy"`
```

I think the semantics of this plus the defaulting in the schema mean that an OG could be submitted without setting upgrade strategy but then when the user fetches it from the cluster, they'd see upgrade strategy populated with the default name.

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -45,6 +45,49 @@ type OperatorGroupSpec struct {
...
+		return DefaultUpgradeStrategy
+	}
+
+	switch {
+	case o.Spec.UpgradeStrategy.Name == UnsafeFailForwardUpgradeStrategy:
+		return o.Spec.UpgradeStrategy.Name
+	default:
+		return DefaultUpgradeStrategy
+	}
 }
```
joelanford, line 91: Based on my above suggestion about making `UpdateStrategy` a non-pointer, I wonder if this function can just be dropped given the defaulting and validation we include in the CRD?

pkg/operators/v1/operatorgroup_types.go
joelanford: Thinking more: I wonder what this would mean for existing OperatorGroups?

pkg/operators/v1/operatorgroup_types.go
awgreene: Sounds reasonable, I only added this method to handle the `nil UpgradeStrategy` case.

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -45,6 +45,49 @@ type OperatorGroupSpec struct {
...
+// replacing phase from the succeeded phase or from the failed phase.
+// Additionally, OLM will generate new installPlans when a subscription references
+// a failed installPlan and the catalog has been updated with a new upgrade for
+// the existing set of operators. This upgrade strategy is unsafe and may result
+// in unexpected behavior or unrecoverable data loss unless you have deep
+// understanding of the set of operators being managed in the namespace.
+type UpgradeStrategy struct {
+	// +kubebuilder:validation:Enum=Default;UnsafeFailForward
+	// +kubebuilder:default=Default
+	Name upgradeStrategyName `json:"name"`
```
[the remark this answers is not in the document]
awgreene: The plan was for customers to use the exported constants, but it should be fine given the enum values defined in the CRD.

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -45,6 +45,49 @@ type OperatorGroupSpec struct {
...
+// allow operators to move to the next version if an installation or upgrade
+// has failed.
+//
+// UnsafeFailForward: OLM will allow clusterServiceVersions to move to the
+// replacing phase from the succeeded phase or from the failed phase.
+// Additionally, OLM will generate new installPlans when a subscription references
+// a failed installPlan and the catalog has been updated with a new upgrade for
+// the existing set of operators. This upgrade strategy is unsafe and may result
+// in unexpected behavior or unrecoverable data loss unless you have deep
+// understanding of the set of operators being managed in the namespace.
```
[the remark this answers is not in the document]
awgreene: +1

pkg/operators/v1/operatorgroup_types.go
awgreene: This change would invalidate OperatorGroups where the spec is defined like so:
```bash
$ k apply -f-<<EOF
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  generation: 3
  name: foo
  namespace: bar
spec: {}
EOF
error: error validating "STDIN": error validating data: ValidationError(OperatorGroup.spec): missing required field "upgradeStrategy" in com.coreos.operators.v1.OperatorGroup.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

Marking this struct as optional allowed the object above to be created but did not generate the default value for the `og.spec.upgradeStrategy.Name` field:
```bash
$ apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"generation":3,"name":"foo","namespace":"bar"},"spec":{}}
  creationTimestamp: "2022-04-05T18:50:07Z"
  generation: 1
  name: foo
  namespace: bar
  resourceVersion: "33408"
  uid: e0d101e9-fc6c-4664-9e09-5fa37372a693
spec: {}
status:
  lastUpdated: "2022-04-05T18:50:07Z"
  namespaces:
  - ""
 ```
We can still make the field a non-pointer, but we would probably want to provide a helper function that treats an unset `og.spec.upgradeStrategy.Name` as the `Default` upgrade strategy.

File: pkg/operators/v1/operatorgroup_types.go (hunk 2 above)
Comment (awgreene) [reply pkg/operators/v1/operatorgroup_types.go]: Unfortunately, due to the reasoning described [here](https://github.com/operator-framework/api/pull/235#discussion_r843161019), we can't rely on the CRD defaulting to set the correct values.

File: pkg/operators/v1/operatorgroup_types.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (joelanford) [reply pkg/operators/v1/operatorgroup_types.go:75]: Correct, but if I want to write a function includes the value of `Name` as a function parameter or return value, I would need to use `upgradeStrategyName`, but I can't because its unexported.

File: pkg/operators/v1/operatorgroup_types.go (hunk 1 above)
Comment (awgreene) [reply pkg/operators/v1/operatorgroup_types.go]: This was inaccurate and has been updated in the latest version of the PR.

File: pkg/operators/v1/operatorgroup_types.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (awgreene) [reply pkg/operators/v1/operatorgroup_types.go:75]: Ah, that's fair. I'll convert it to a string.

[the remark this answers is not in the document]
Comment (njhale) [reply pkg/operators/v1/operatorgroup_types.go]: Rather than string, I think this should be a named, exported type.

**Review by njhale — COMMENTED** [review]
Thanks for adding this -- a few questions/comments:

File: pkg/operators/v1/operatorgroup_types.go [hunk 5]
```diff
@@ -45,9 +45,32 @@ type OperatorGroupSpec struct {
...
+	// TechPreviewUnsafeFailForward: OLM will allow clusterServiceVersions to move to the
+	// replacing phase from the succeeded phase or from the failed phase.
+	// Additionally, OLM will generate new installPlans when a subscription references
+	// a failed installPlan and the catalog has been updated with a new upgrade for
+	// the existing set of operators.
+	//
+	// WARNING: The TechPreviewUnsafeFailForward upgrade strategy is unsafe and may result
+	// in unexpected behavior or unrecoverable data loss unless you have deep
+	// understanding of the set of operators being managed in the namespace.
+	// OperatorGroupStatus is the status for an OperatorGroupResource.
```
timflannagan, line 66: Looks like this got lot in the most recent changes:

```go
	// OperatorGroupStatus is the status for an OperatorGroupResource.
```

Is in the UpgradeStrategy godoc vs. above the OperatorGroupStatus type.

awgreene: Oops, will fix

exdx (approved):

pkg/operators/v1/operatorgroup_test.go
```diff
@@ -0,0 +1,61 @@
...
+					UpgradeStrategy: "foo",
+				},
+			},
+			expected: UpgradeStrategyDefault,
+		},
+		{
+			description: "UnsafeFailForwardUpgradeStrategy",
+			og: &OperatorGroup{
+				Spec: OperatorGroupSpec{
+					UpgradeStrategy: "TechPreviewUnsafeFailForward",
```
timflannagan, line 49: nit: use the constant here?

pkg/operators/v1/operatorgroup_types.go
```diff
@@ -45,6 +64,29 @@ type OperatorGroupSpec struct {
...
+	// the existing set of operators.
+	//
+	// WARNING: The TechPreviewUnsafeFailForward upgrade strategy is unsafe and may result
+	// in unexpected behavior or unrecoverable data loss unless you have deep
+	// understanding of the set of operators being managed in the namespace.
+	//
+	// +kubebuilder:validation:Enum=Default;TechPreviewUnsafeFailForward
+	// +kubebuilder:default=Default
+	// +optional
+	UpgradeStrategy string `json:"upgradeStrategy,omitempty"`
```
timflannagan, line 89: I think @njhale might've mentioned this but should we have a dedicated string type for the UpgradeStrategy here?

timflannagan (approved):

pkg/operators/v1/operatorgroup_types.go
awgreene: Sure.

awgreene: Created a string alias named `UpgradeStrategy`.

pkg/operators/v1/operatorgroup_test.go
awgreene: This acts as a warning if the constant changes value which is a backwards incompatible change.

PR #238, opened 2022-04-14
✨ change: move check bundle name to the Good Practices validator

camilamacedo86:
**Description**
change: move the check bundle name to the Good Practices validator

**Motivation**
Now, that we have the Good Practices one the warning when the bundle name when is not following the semantics `<operator-name>.v<semver>` shows fit better in this category of checks.


pkg/validation/internal/good_practices.go
```diff
@@ -96,6 +104,34 @@ func validateResourceRequests(csv *operatorsv1alpha1.ClusterServiceVersion) (err
 	return errs, warns
 }
 
+// checkBundleName will validate the operator bundle name informed via CSV.metadata.name.
+// The motivation for the following check is to ensure that operators authors knows that operator bundles names should
+// follow a name and versioning convention
+func checkBundleName(csv *operatorsv1alpha1.ClusterServiceVersion) []error {
+	var warns []error
+	// Check if is following the semver
+	re := regexp.MustCompile("([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z-]+)?$")
```
grokspawn, line 113: Is there a reason that we're not using the regex from [semver](https://semver.org/)?
here:
> "([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z-]+)?$"
there:
> "^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"

This one does not work for prerelease versions, right?

camilamacedo86: The go here is silly. Only move to the validator that it fits better

grokspawn: Opened #246.  It shouldn't block this, anyway. 

/lgtm

dinhxuanvu (approved):
/approve

PR #241, opened 2022-05-14
:sparkles: (GoodPracticesValidator) : Add a new check to warn authors when permissions to create CRDs are found

camilamacedo86:
**Description**
Add a new check to warn authors when permissions to create CRDs are found. They usually will ask for these permissions when their Operands manage other operators and create CRDs that is not recommended. 

**Motivation**
Use case: we recently check a Helm-based language operator created with helm charts which would install other Operators.  

**How it will be used**

Any project that uses OF/api will have this check. However, via SDK (which is generally used by users/ci's/pipelines) the warning will be raised when:

a) To validate a bundle against the validator for Good Practices specifically, in addition to required bundle validators: `$ operator-sdk bundle validate ./bundle --select-optional name=good-practices`

b) To validate a bundle against the entire suite of validators for Operator Framework, in addition to required bundle validators: `$ operator-sdk bundle validate ./bundle --select-optional suite=operatorframework`

### DIFF
--- a/pkg/validation/internal/good_practices.go
+++ b/pkg/validation/internal/good_practices.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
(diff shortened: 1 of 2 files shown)


camilamacedo86:
c/c @ryantking

pkg/validation/internal/good_practices.go
```diff
@@ -116,6 +130,68 @@ func validateHubChannels(channels []string) error {
...
+										}
+									}
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
ryantking, line 190: I think that we can simplify this using `k8s.io/apimachinery/pkg/util/sets`:

```suggestion
    verbSet := sets.NewString(verbs...)
    for _, rule := range perm.Rules {
    	for group, resources := range apiGroupResourcesMap {
    	    if sets.NewString(rule.APIGroups).Has(group) && 
    	      sets.NewString(resources).Has(rule.Resource) 
    	      verbSet.HasAny(rule.Verbs...) {
    	        return true
            }
        }
    }
```

ryantking: I didn't fully test it so it probably needs to be tweaked a bit.

I think using sets for quick lookups is easier to read than a bunch of nested loops, but its your call. Its a suggestion, not requested changes.

camilamacedo86: I tried to check it, and it did not work well as expected.
Also, I found little hard to understand the code with the proposed changes.
So, do you think that we can move forward without applying this one?
If yes, could you lgtm this PR?

pkg/validation/internal/good_practices.go
```diff
@@ -116,6 +130,64 @@ func validateHubChannels(channels []string) error {
...
+			return warning
+		}
+	}
+
+	return nil
+}
+
+func hasRBACFor(perm v1alpha1.StrategyDeploymentPermissions, apiGroupResourceMap map[string][]string, verbs []string) bool {
+	// For each APIGroup and list of resources that we are looking for
+	for apiFromMap, resourcesFromMap := range apiGroupResourceMap {
```
dinhxuanvu, line 161: I wonder if we have a better way to check for this. This looks rather ugly though I guess you didn't have much of a choice.

dinhxuanvu (commented):
Even though I can see the reason behind this PR, this change will enforce some restrictions on what an operator can do. I would like to run this pass another set of eyes before approving.
@joelanford do you have any issues with this change?

dinhxuanvu:
Also this PR needs a rebase.

pkg/validation/internal/good_practices.go
camilamacedo86: I think it is easily readable and understandable then we try to make it less ugly when behind the scene will do the same.

camilamacedo86:
Hi @dinhxuanvu:

> Even though I can see the reason behind this PR, this change will enforce some restrictions on what an operator can do. I would like to run this pass another set of eyes before approving.
@joelanford do you have any issues with this change?

We are not enforcing. The good practices validator will return a warning in this case and not an error.
It was a @joelanford's suggestion too (More info: https://coreos.slack.com/archives/GHMALGJV6/p1652210479908649?thread_ts=1652185816.258749&cid=GHMALGJV6)

From ^
> We could also include a warning in the bundle validation somewhere that checks to see if the bundle includes RBAC that allows it to create CRDs. And if it does, we could raise a warning pointing to that best practice doc.

dinhxuanvu (approved):
/approve

PR #242, opened 2022-05-19
chore: bump k8s deps to 1.24 & makefile go 1.18 support

everettraven:
## Description of change
- Bump k8s dependencies to 1.24
- Modify Makefile to support Go 1.18

## Motivation for the change
Kubernetes 1.24 has been released and is built with Go 1.18, so we need to update this library to support k8s 1.24 so the rest of the Operator Framework tools can begin making changes to support k8s 1.24. This also modifies the Makefile to allow for the support of Go 1.18 when it comes time to make the Go version bump


go.mod
```diff
@@ -1,23 +1,23 @@
 module github.com/operator-framework/api
 
-go 1.17
+go 1.18
```
camilamacedo86, line 3: I do not think that we should upgrade here. 
That will force all consumers use go 1.18
WDYT about we make all projects support go 1.18 first?

everettraven: That is a good point.

I was thinking it would be best to try and do the k8s 1.24 and Go 1.18 bump at the same time since k8s 1.24 is built with Go 1.18, but now that you point it out it may not be as easy for other projects to make the Go 1.18 bump. 

I think it may be best to split it into 2 separate bumps - one for k8s 1.24 and then one for Go 1.18. I will go ahead and update this. 

Thanks for pointing this out @camilamacedo86

rashmigottipati:
I did mention to Bryce we should be bumping 1.18 as k8s 1.24 is built using Go 1.18. 
@everettraven however, I agree with Camila's point, so I'm good with switching to use go 1.17 here.

rashmigottipati (approved):
/lgtm

camilamacedo86 (approved):
That shows fine 🥇 

/approved

dinhxuanvu (commented):
I don't see any problems with this PR specifically. However, given registry and olm repos do vendor api, these two repos will need to do a k8s 1.24 bump as well. Registry is still in 1.22 while olm is on 1.23 already. Assuming they have no problems with 1.24 bump, then this PR should be good to go.
@joelanford FYI

njhale:
`go-apidiff` is complaining about `ClusterName` being deprecated. IIUC, it wasn't used outside of KCP upstream anyway.

/override Go/go-apidiff

everettraven:
As per @njhale 

/override go-apidiff

everettraven:
> I don't see any problems with this PR specifically. However, given registry and olm repos do vendor api, these two repos will need to do a k8s 1.24 bump as well. Registry is still in 1.22 while olm is on 1.23 already. Assuming they have no problems with 1.24 bump, then this PR should be good to go. @joelanford FYI

@dinhxuanvu Are you suggesting that OLM and Registry are updated to support k8s 1.24 before this PR is merged?

PR #256, opened 2022-08-18
pkg/metadata: bundle metadata utils and types

bcrochet:
From the original PR:

There should be a canonical set of utilities and types that can find and parse an annotations.yaml in a bundle directory instead of re-implementing that logic in various ways in various operator-framework repos. This repo is a good home for them because it is the source of truth for many OF APIs.

/kind feature

Rebase of #60


perdasilva:
closing as stale

PR #257, opened 2022-08-25
(catsrc) introduce spec.runAsRoot field

anik120:
With the [change](https://github.com/operator-framework/operator-registry/pull/974) in opm being copied to a /tmp
folder rather than / (root), a registry pod created in a namespace labled
enforce:restricted for the Pod Security Admission controller is created by
the catalog operator with the (appropriate securityContext details) https://github.com/operator-framework/operator-lifecycle-manager/pull/2820/files#diff-fffdeef1fc140a5dc5dc92dda323f567a6e46fc2ecbb0b91ba907acd02bde50dR185-R210
to run it in restricted mode.

However, Catalogs built with a version of opm that does not contain the above
change still needs privileged permission to run in a namespace that has to
be labeled as enforce:privileged for the PSA controller.

This PR introduces a new field, spec.runAsRoot, so that admins can indiciate
their intent to allow to run the old CatalogSource in a privileged mode.
When the catalog operator sees this field set to true, it will not set the
securityContext in the registry pod to `runAsNonRoot:true`. Instead, it will
set the securityContext to `runAsNonRoot:false`.


anik120:
closing this in favor of #258 since the `make manifests` target is running into issues for me (and potentially broken for all mac users). Will have to investigate after this work is done with.

PR #261, opened 2022-09-08
(catsrc) rename grpcPodConfig.runAsRoot to grpcPodConfig.SecurityContextConfig

anik120:
Accepted values for SecurityContextConfig(string): legacy/restricted(default)


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -113,10 +113,11 @@ type GrpcPodConfig struct {
...
 	PriorityClassName *string `json:"priorityClassName,omitempty"`
 
-	// RunAsRoot allows admins to indicate that they wish to run the container inside the CatalogSource pod in a privileged
-	// mode as root.  This should only be enabled when running older catalog images which could not be run as non-root.
+	// RequirePrivilegeAccess when set to true, allows the containers in the pod to run in privileged mode.
+	// This should only be set to true when using older catalog images, which could not be run in restricted mode,
+	// and require privileged access.
 	// +optional
-	RunAsRoot bool `json:"runAsRoot,omitempty"`
+	RequirePrivilegedAccess bool `json:"runAsRoot,omitempty"`
```
dinhxuanvu, line 120: Should the json tag be `RequirePrivilegedAccess` as well?

dinhxuanvu (changes requested):

pkg/operators/v1alpha1/catalogsource_types.go
anik120: oof nice catch, thanks @dinhxuanvu!

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -79,9 +79,9 @@ spec:
...
-                      type: boolean
+                    securityContextConfig:
+                      description: 'SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource''s pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) controller''s `restricted` mode, or doesn''t set these values at all, in which case the pod can only be run in PSA `privileged` namespaces. By default, SecurityContextConfig is set to `restricted`. If the value is unspecified, the default value of `restricted` is used.  Specifying any other value will result in a validation error. When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to `legacy`. More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'''
```
grokspawn, line 83:
```suggestion
                      description: 'SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource's pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) controller's `restricted` mode, or doesn't set these values at all, in which case the pod can only be run in PSA `privileged` namespaces. By default, SecurityContextConfig is set to `restricted`. If the value is unspecified, the default value of `restricted` is used.  Specifying any other value will result in a validation error. When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to `legacy`. More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'
```

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -113,10 +120,45 @@ type GrpcPodConfig struct {
...
-	RunAsRoot bool `json:"runAsRoot,omitempty"`
+	SecurityContextConfig SecurityConfig `json:"securityContextConfig,omitempty"`
+}
+
+// UnmarshalJSON for GrpcPodConfig is a custom unmarshaler that
+// sets the default value of `restricted` for SecurityContextConfig
+// if a value isn't set explicitly, and throws an error if an incorrect
+// value (a value other than `restricted` or `legacy` ) is set for
+// SecurityContextConfig
+func (g *GrpcPodConfig) UnmarshalJSON(data []byte) (err error) {
```
joelanford, line 139: Haven't we run into major issues with relying on `UnmarshalJSON`-based defaulting and validating? IIRC, we had a problem with the catalog source polling interval duration value where an invalid catsrc existing in etcd caused the entire informer to fail, making it impossible to reconcile all catalog sources.

Can we instead use CRD schema defaulting and validating?

anik120: @joelanford done.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -113,10 +120,17 @@ type GrpcPodConfig struct {
...
+	// the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in
+	// Pod Security Admission(PSA) "restricted" mode, or does not set these values at all, in which case the pod
+	// can only be run in PSA "privileged" namespaces. By default, SecurityContextConfig is set to "restricted". If the value
+	// is unspecified, the default value of restricted is used.  Specifying any other value will result in a validation error.
+	// When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to
+	// "legacy". More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/
 	// +optional
-	RunAsRoot bool `json:"runAsRoot,omitempty"`
+	// +kubebuilder:validation:Enum=legacy;restricted
+	// +kubebuilder:default:=restricted
```
joelanford, line 132: Not positive it matters (or if my suggestion is even correct) since it looks like the CRD generated correctly.
```suggestion
	// +kubebuilder:default=restricted
```

awgreene: According to the [official Kubebuilder Docs](https://book.kubebuilder.io/reference/markers/crd-validation.html#crd-validation), this is correct syntax.

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -113,10 +120,17 @@ type GrpcPodConfig struct {
...
 
-	// RunAsRoot allows admins to indicate that they wish to run the container inside the CatalogSource pod in a privileged
-	// mode as root.  This should only be enabled when running older catalog images which could not be run as non-root.
+	// SecurityContextConfig can be one of "legacy" or "restricted". The grpc pod created is either injected with
+	// the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in
+	// Pod Security Admission(PSA) "restricted" mode, or does not set these values at all, in which case the pod
+	// can only be run in PSA "privileged" namespaces. By default, SecurityContextConfig is set to "restricted". If the value
+	// is unspecified, the default value of restricted is used.  Specifying any other value will result in a validation error.
+	// When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to
+	// "legacy". More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/
```
awgreene, line 129: Documentation Nit, I felt that some of the information concerning possible values to be apparent through the schema validation changes Anik recently introduced. Can be ignored if you don't think value is added @anik120 
```suggestion
	// SecurityContextConfig allows users to configure a grpc pod's pod.spec.securityContext and pod.spec.container[*].securityContext so the pod may meed supported Pod Security Admission(PSA) modes.
	// By default, the grpc pod will be ran in "restricted" mode.
	// If set to "legacy" mode, the grpc pod's pod.spec.securityContext and pod.spec.container[*].securityContext
	// will not be set and the pod will only run in PSA "privileged" namespaces. In most cases, "legacy" mode
	// should be used with older catalog images, which could not be run in restricted mode, 
	// More information about PSA can be found here: 
	// https://kubernetes.io/docs/concepts/security/pod-security-admission/
```

awgreene (changes requested):
More or less ready to approve this. Great work @anik120!

pkg/operators/v1alpha1/catalogsource_types.go
anik120: @awgreene changed the doc, with slight modifications. PTAL!

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -79,9 +79,13 @@ spec:
...
                       type: string
-                    runAsRoot:
-                      description: RunAsRoot allows admins to indicate that they wish to run the container inside the CatalogSource pod in a privileged mode as root.  This should only be enabled when running older catalog images which could not be run as non-root.
-                      type: boolean
+                    securityContextConfig:
+                      description: 'SecurityContextConfig allows users to configure a grpc pod''s pod.spec.securityContext and pod.spec.container[*].securityContext to meet expected Pod Security Admission(PSA) standards. By default, the grpc pod is configured to run in "restricted" mode. If set to "legacy" mode, the grpc pod''s pod.spec.securityContext and pod.spec.container[*].securityContext will not be set, and the pod will only run in PSA "privileged" namespaces. The "legacy" mode should be used only with older catalog images, which could not be run in restricted mode, More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'
```
perdasilva, line 83: don't mean to throw a spanner in the works, but can it also run under 'baseline'?

anik120: Yea looks like you're right. It can in fact run in Baseline. Updating the doc.

anik120: @perdasilva done.

pkg/operators/v1alpha1/catalogsource_types.go
awgreene: LGTM

PR #266, opened 2022-10-05
transplanting declcfg read/write to api

grokspawn:
operator-registry side of this transaction is https://github.com/operator-framework/operator-registry/pull/1030


go.mod
```diff
@@ -7,6 +7,9 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-bindata/go-bindata/v3 v3.1.3
 	github.com/google/cel-go v0.12.4
+	github.com/h2non/filetype v1.1.3
+	github.com/h2non/go-is-svg v0.0.0-20160927212452-35e8c4b0612c
+	github.com/joelanford/ignore v0.0.0-20210610194209-63d4919d8fb2
```
anik120, line 12: We should probably bring the ignore package in here instead of importing it from Joe's, doesn't look like it'll be too much work: https://github.com/joelanford/ignore/blob/main/ignore.go

anik120 (commented):
So the transplant looks good overall, but my only concern is the API support we are taking on by bringing this here from operator-registry/alpha/. 

Do we have a list of projects other than operator-registry that'll be using the `declcfg` package from here? Should we put all of these inside an `alpha` folder instead so that it's communicated that everything in this package is subject to change without heavy notifications?

grokspawn:
> So the transplant looks good overall, but my only concern is the API support we are taking on by bringing this here from operator-registry/alpha/.
> 
> Do we have a list of projects other than operator-registry that'll be using the `declcfg` package from here? Should we put all of these inside an `alpha` folder instead so that it's communicated that everything in this package is subject to change without heavy notifications?

I definitely agree with preserving the alpha here, and it was just sheet oversight that had me drop it. :( 

There is no current list of projects using `declcfg`.  This move is to facilitiate potential development on a shared base.

anik120:
Okay this /lgtm 👍🏽

anik120:
/lgtm 

I suppose the command has to be in a separate line

go.mod
anik120: Would be great to create an issue before we merge this PR so that we don't loose track of this

joelanford: The reason this is separate is because `openshift/oc` needs it too, and `oc` does not import `operator-framework` stuff (or at least isn't currently).

If we move that ignore package here, then we need to do one of the following:
1. Maintain the code separately for `oc` somewhere.
2. Get `oc` to agree to require this API repo.

Thread about it here: https://github.com/openshift/oc/pull/868/files#r668149063

pkg/lib/alpha/declcfg/declcfg.go
```diff
@@ -0,0 +1,109 @@
+package declcfg
```
joelanford, line 1: Pet peave/nit: What's the deal with `pkg/lib`? In Go `pkg` is sort of the convention of where to put library code in a project that _also_ has a main executable.

It has always been odd to me that we have this structure in this repo.

grokspawn: I had that question as well, but sought to 'when in Rome' the problem.  
If you feel strongly, I can move all this code to pkg/alpha/fbc instead of pkg/lib/alpha/fbc.  I didn't have enough historical context for a good position.

grokspawn: Moved.

grokspawn:
ostensibly we still want to do this 'some day', but there's no benefit (and much cost) to trying to adapt this PR for a future effort.  Closed will improve signal.

PR #273, opened 2022-12-01
Default to legacy psa settings

awgreene:
The catalogSource changes that introduced restricted pod security admission settings by default were rolled out a bit too hastily and have not provided users with enough time to adjust to the changes. This commit sets the default catalogSource config changes to "legacy" mode by default.

Signed-off-by: Alexander Greene <greene.al1991@gmail.com>


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -123,14 +123,14 @@ type GrpcPodConfig struct {
 	// SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource's pod is either injected with
 	// the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in
 	// Pod Security Admission(PSA) controller's `restricted` mode, or doesn't set these values at all, in which case the pod
-	// can only be run in PSA `baseline` or `privileged` namespaces. By default, SecurityContextConfig is set to `restricted`.
-	// If the value is unspecified, the default value of `restricted` is used.  Specifying any other value will result in a
+	// can only be run in PSA `baseline` or `privileged` namespaces. By default, SecurityContextConfig is set to `legacy`.
+	// If the value is unspecified, the default value of `legacy` is used.  Specifying any other value will result in a
```
anik120, line 127: We should probably also add that "the default will be switched to `restricted` very soon and if you are a catalog author then please ensure that your next catalog image release is compatible with the `restricted` mode"

awgreene: Done!

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -80,9 +80,9 @@ spec:
...
+                      description: "SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource's pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) controller's `restricted` mode, or doesn't set these values at all, in which case the pod can only be run in PSA `baseline` or `privileged` namespaces. By default, SecurityContextConfig is set to `legacy`. If the value is unspecified, the default value of `legacy` is used. Specifying a value other than `legacy` or `restricted` result in a validation error. When using older catalog images, which could not be run in `restricted` mode, the SecurityContextConfig should be set to `legacy`. \n The default will soon be set to `restricted`, catalog maintainers should rebuild their catalogs with a version of opm that supports running catalogSource pods in `restricted` mode to prepare for these changes. \n More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'"
```
bparees, line 83:
```suggestion
                      description: "SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource's pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) `restricted` mode, or doesn't set these values at all, in which case the pod can only be run in PSA `baseline` or `privileged` namespaces. Currently if the SecurityContextConfig is unspecified, the default value of `legacy` is used. Specifying a value other than `legacy` or `restricted` result in a validation error. When using older catalog images, which could not be run in `restricted` mode, the SecurityContextConfig should be set to `legacy`. \n In a future version will the default will be set to `restricted`, catalog maintainers should rebuild their catalogs with a version of opm that supports running catalogSource pods in `restricted` mode to prepare for these changes. \n More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'"
```

awgreene: Done

awgreene:
@bparees @anik120 this PR is ready for a second review.

bparees:
lgtm

PR #274, opened 2022-12-12
(bugfix): make k8s 1.25 validation logic check api group before issuing a warning

everettraven:
## Description of Change
This PR updates the k8s 1.25 validation logic and the associated test cases to verify if any of the `apiGroups` defined in an RBAC `PolicyRule` has a potential deprecation. If it finds one it will check a combination of the `apiGroup` and all resources defined in the `PolicyRule` (i.e `apiGroup/resource`) to see if any of these combinations results in a deprecated API - if so it returns a warning.

## Motivation for Change
There was a bug in the validation logic where it was issuing a warning even if the API group the resource belonged to had not been deprecated.

### DIFF
--- a/pkg/validation/internal/removed_apis.go
+++ b/pkg/validation/internal/removed_apis.go
@@ -12,6 +12,7 @@ import (
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // k8sVersionKey defines the key which can be used by its consumers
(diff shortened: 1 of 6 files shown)


pkg/validation/internal/removed_apis.go
```diff
@@ -403,11 +435,19 @@ func getRemovedAPIsOn1_25From(bundle *manifests.Bundle) (map[string][]string, ma
...
-									if _, ok := resInCsvCrds[res]; ok {
-										continue
+								for _, apiGroup := range rule.APIGroups {
+									// if the API Group being used has a potential deprecation
+									// lets check if it is using a deprecated resource as well
+									if apiGroupHasDeprecation(apiGroup) {
+										// check a combination of apiGroup/resource for all
+										// the resources to see if any match the deprecated apis
+										for _, res := range rule.Resources {
+											if _, ok := resInCsvCrds[res]; ok {
```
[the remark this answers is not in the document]
joelanford: I think that is actually a check that’s saying “if this resource is already covered by one that’s listed in the CSV as something the operator uses, then that’s a more reliable check, so skip this check”

This might be something we want to add a comment for since it isn't immediately obvious why we'd be bailing out when we find an RBAC rule for a resource within a deprecated API.

pkg/validation/internal/removed_apis.go
```diff
@@ -335,25 +336,56 @@ func getRemovedAPIsOn1_25From(bundle *manifests.Bundle) (map[string][]string, ma
...
-			warnDeprecatedAPIs[res] = append(warnDeprecatedAPIs[res], msg)
-		case "horizontalpodautoscalers":
-			warnDeprecatedAPIs[res] = append(warnDeprecatedAPIs[res], msg)
-		case "poddisruptionbudgets":
-			warnDeprecatedAPIs[res] = append(warnDeprecatedAPIs[res], msg)
-		case "podsecuritypolicies":
-			warnDeprecatedAPIs[res] = append(warnDeprecatedAPIs[res], msg)
-		case "runtimeclasses":
-			warnDeprecatedAPIs[res] = append(warnDeprecatedAPIs[res], msg)
+	warnIfDeprecated := func(gr schema.GroupResource, msg string) {
```
joelanford, line 339: I wonder if there's a way to have a `map[schema.GroupResource]struct{}` that is essentially a set of the group/resources that have deprecations in their ranks.

It seems like that could cleanup a lot of the duplication I'm seeing here.

Similar comment for the GVKs in `addIfDeprecated`.

pkg/validation/internal/removed_apis.go
```diff
@@ -335,25 +336,56 @@ func getRemovedAPIsOn1_25From(bundle *manifests.Bundle) (map[string][]string, ma
...
+				warnDeprecatedAPIs[gr.Resource] = append(warnDeprecatedAPIs[gr.Resource], msg)
+			}
+		case "node.k8s.io":
+			if gr.Resource == "runtimeclasses" {
+				warnDeprecatedAPIs[gr.Resource] = append(warnDeprecatedAPIs[gr.Resource], msg)
+			}
 		}
 	}
 
+	apiGroupHasDeprecation := func(group string) bool {
```
joelanford, line 368: Same question here. Can we make this a `map[string]struct{}` (or use `sets.String` from k8s.io/apimachinery)?

theishshah (approved):

pkg/validation/internal/removed_apis.go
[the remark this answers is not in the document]
everettraven: Taking a look at this, I think it could really go either way. The intention behind this section is exactly what @joelanford mentioned.

I think we could either add a comment or adopt the recommendation by @oceanc80 and remove the `continue` statement within the block.

i.e go from:
```go
for _, res := range rule.Resources {
	if _, ok := resInCsvCrds[res]; ok {
		continue
	}
       warnIfDeprecated(schema.GroupResource{Group: apiGroup, Resource: res}, fmt.Sprintf("ClusterServiceVersion.Spec.InstallStrategy.StrategySpec.%s[%d].Rules[%d]", permField, i, j))
}
```
to:
```go
for _, res := range rule.Resources {
	if _, ok := resInCsvCrds[res]; !ok {
            warnIfDeprecated(schema.GroupResource{Group: apiGroup, Resource: res}, fmt.Sprintf("ClusterServiceVersion.Spec.InstallStrategy.StrategySpec.%s[%d].Rules[%d]", permField, i, j))
	}
}
```

Technically saves us a line of code and if it feels more readable that way I am happy to change it

pkg/validation/internal/removed_apis.go
everettraven: This should be resolved as of: https://github.com/operator-framework/api/pull/274/commits/f31e203a46d8e8b7bd517952fe927865365ad2d9

pkg/validation/internal/removed_apis.go
everettraven: After adding maps to resolve your previous comment, I ended up removing this function definition as the map used by `warnIfDeprecated` has the `schema.GroupResource` as a key. If the given `schema.GroupResource` doesn't match any of the keys it doesn't warn.

I think the only downside to having removed this function is that now it always runs the loop to create a `schema.GroupResource` and verify it instead of only if the group has a deprecated api. If this is something we are concerned about I can re-implement this function.

@joelanford @oceanc80 WDYT?

PR #276, opened 2023-01-28
Update multi-arch validator to check node affinity for multi-platform environments

jaypoulz:
Adding a validator to image references for nodeAffinity specifications.
Will check for the following for the CSV images:
1. Missing node affinity boundaries
2. Incomplete node affinity boundaries (image supports more than node affinity)
3. Extra node affinity settings that are not supported by the image data

In doing so I've built upon the work done by @camilamacedo86.

### DIFF
--- a/pkg/validation/internal/multiarch.go
+++ b/pkg/validation/internal/multiarch.go
@@ -10,6 +10,9 @@ import (
 	"github.com/operator-framework/api/pkg/manifests"
 	"github.com/operator-framework/api/pkg/validation/errors"
 	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+	corev1 "k8s.io/api/core/v1"
+
+	log "github.com/sirupsen/logrus"
 )
 
 // MultipleArchitecturesValidator validates the bundle against criteria to support Multiple Architectures. For further
@@ -26,11 +29,13 @@ import (
 //
 // ### What is checked?
 //
-// On this check, we aggregate the archetype(s) and OS(s) provided via the labels and those which are found by checking the images so that, we can check:
+// On this check, we aggregate the platform architecture(s) and OS(s) provided via the labels and those which are found by checking the images so that, we can check:
 //
 // - If your CSV is missing labels
 //
-// - If your Operator bundle specifies images which do not support all archetypes found for your Operator image(s) (probably supported by your project)
+// - If your Operator bundle specifies images which do not support all architectures found for your Operator image(s) (probably supported by your project)
+//
+// - If your deployment spec follows the best practice of setting nodeAffinity to ensure image(s) are only scheduled on compatible platform nodes.
 //
 // Note: To better guess the case scenarios where authors might have missed the labels, the following check will verify all architectures supported by the Operator image(s). However, by looking at the CSV we are not able to ensure what is the Operator image because this info is not provided. Therefore, we know by SDK the Operator image container will be called manager.
 //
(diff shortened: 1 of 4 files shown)


jaypoulz:
/assign awgreene
I believe that an extension to this which also scans for node affinity discrepancies in the rest of the bundle is desirable, but I wanted to get this patch up for early feedback.

gallettilance (approved):
Looks good! Don't have much to say - everything is clear and tests are extensive.

pkg/validation/internal/multiarch.go
```diff
@@ -205,58 +218,123 @@ func (data *multiArchValidator) loadInfraLabelsFromCSV() {
...
+		return platforms
+	}
+
+	var terms = spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	for _, t := range terms {
+		var arches = make([]string, 0)
+		var os = make([]string, 0)
+
+		for _, e := range t.MatchExpressions {
+			if e.Operator != "In" {
```
awgreene, line 313: This won't support anti-affinity, which I think is fine given the goal of this change.

pkg/validation/internal/multiarch.go
```diff
@@ -297,12 +375,12 @@ func (data *multiArchValidator) inspectImages(images map[string][]platform) map[
 			if err != nil {
 				data.warns = append(data.warns, fmt.Errorf("unable to inspect the image (%s) : %s", k, err))
 
-				// We set the Arch and SO as error for we are able to deal witth these cases further
+				// We set the Arch and OS as error for we are able to deal witth these cases further
```
awgreene, line 378:
```suggestion
				// We set the Arch and OS as error for we are able to deal with these cases further
```

pkg/validation/internal/multiarch.go
```diff
@@ -325,21 +403,27 @@ func (data *multiArchValidator) doChecks() {
...
+	data.checkNodeAffinity(data.otherCSVDeploymentImages)
 }
 
 // checkMissingSupportForOtherImages checks if any image is missing some arch or so found
-// (probably the image should support the arch or SO )
-func (data *multiArchValidator) checkMissingSupportForOtherImages() {
-	for image, plaformFromImage := range data.allOtherImages {
+// (probably the image should support the arch or OS )
+func (data *multiArchValidator) checkMissingSupportForOtherImages(images map[string][]platform) {
+	for image, platformFromImage := range images {
```
awgreene, line 421: Note: No functional changes were made in this function other than the new argument.

pkg/validation/internal/multiarch.go
```diff
@@ -384,27 +468,128 @@ func (data *multiArchValidator) checkMissingSupportForOtherImages() {
...
+
+		// Verify we were able to gather valid platform data for the image
+		imagePlatformDataValid := true
+		for _, imageData := range platformFromImage {
+			// Ignore the case when the Platform.Architecture == "error" since that means
+			// that was not possible to inspect the image
+			if imageData.Architecture == "error" {
+				imagePlatformDataValid = false
+				break
+			}
```
awgreene, line 501: Could you explain why it is okay not to log an error here? Is it logged elsewhere when the image fails to be inspected?

awgreene (commented):
Looks good, but I'd like to confirm that warnings do not prevent an operator from being published. Thanks for the  contribution.

jaypoulz:
Some notes on this PR since I've been evaluating it via the operator-sdk instructions as well as trying to run it through the operator-framework/audit tool:

The audit report is going to flag all operators as those with warnings since by default, the operator scaffolding does not provide a kube-rbac-proxy node affinity specification. I think this may need to be updated as part of https://github.com/operator-framework/operator-sdk/pull/5983 (or another PR focused on providing more opinionated scaffolding). While this report is technically accurate, it does muddy the signal for the 40 or so operators in the redhat-operator-index that have otherwise been compliant as per Camila's original report. I don't think we should remove it or skip that image, but I do think that it would be good to add nodeAffinity for this image to the scaffolding so future operators don't hit this.

pkg/validation/internal/multiarch.go
jaypoulz: The multiarch validator is used in the operator-framework/audit tool to produce a report on the health of operators and is also available for self-validation for operator-sdk authors. It will not block operator publishing. :)

pkg/validation/internal/multiarch.go
jaypoulz: It is - specifically it is logged at inspection time in line 382.
https://github.com/operator-framework/api/pull/276/files#diff-20ba073b3d34ec5ee84564e23371a9a0653ca0528bd97fe76837efa1289d8c88R382

dtfranz:
/lgtm
Thanks for taking this on!

PR #283, opened 2023-04-25
Restores openapi-gen tag for `APIResourceReference`

m1kola:
It looks like it was accidentally removed in #168.

Now when we [update `github.com/operator-framework/api` package](https://github.com/operator-framework/operator-lifecycle-manager/pull/2954) for OLM, packageserver fails with the following error:

> unable to get openapi models: cannot find model definition for github.com/operator-framework/api/pkg/operators/v1alpha1.APIResourceReference. If you added a new type, you may need to add +k8s:openapi-gen=true to the package or type and run code-gen again


ankitathomas (approved):
/lgtm

PR #291, opened 2023-08-18
OCPBUGS-17791: CatalogSource: update GoDoc

stevekuznetsov:
We won't be setting a memory limit, so udpate the doc.

Impl: https://github.com/operator-framework/operator-lifecycle-manager/pull/3015


perdasilva (approved):
Lgtm

awgreene:
/hold
Before we proceed with merging this PR, I want to have a conversation on the backwards incompatible change introduced in this PR. From [the related OLM PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/3015), it seems like we're introducing this API change to meet an OpenShift core payload requirement. While some catalogSources are defaults on cluster, users can and do create their own catalogSources.

**Are we certain that these API changes won't impact existing user workflows on custom catalogSources?** IIRC:
- a pod that exceeds it's memory limit will be killed.
- Memory granted to pods isn't yielded back even if the pod doesn't actively need the memory (For example, if the pod needs a bunch of memory at startup it won't abandon it later even if it requires less resources).

So if a user had been using this field, the pod would crash loop assuming that it always requires the same memory that initially caused the crash.

PR #293, opened 2023-09-07
catalogsource: allow configuring gRPC server to use well-known opm

stevekuznetsov:
As the file-based catalog schemas mature, the couping between `opm` and the catalog data is effectively removed. We can take advantage of this new property in the system to not require that a server binary exist in the catalog index image, and remove many of the pain-points that come from using the embedded server binary.

part of https://github.com/operator-framework/operator-lifecycle-manager/issues/3019

/assign @joelanford


pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -155,6 +155,11 @@ type GrpcPodConfig struct {
 	//
 	// This field is best-effort; if unset, no default will be used and no Pod memory limit or $GOMEMLIMIT value will be set.
 	MemoryTarget *resource.Quantity `json:"memoryTarget,omitempty"`
+
+	// UseWellKnownServer configures the gRPC catalog Pod to use a well-known version of the `opm` server. The catalog
+	// index image that this CatalogSource is configured to use *must* be using the file-based catalogs in order to
+	// utilize this feature.
+	UseWellKnownServer bool `json:"useWellKnownServer,omitempty"`
```
joelanford, line 162: What made you move away from the proposed API in https://github.com/operator-framework/operator-lifecycle-manager/issues/3019?

```yaml
    extractContent:
      cacheDir: /var/cache/
      configDir: /var/config/
```

pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -154,7 +154,22 @@ type GrpcPodConfig struct {
...
+	// +optional
+	ExtractContent *ExtractContentConfig `json:"extractContent,omitempty"`
+}
+
+// ExtractContentConfig configures context extraction from a file-based catalog index image.
+type ExtractContentConfig struct {
+	// CacheDir is the directory storing the pre-calculated API cache.
+	CacheDir string `json:"cacheDir,omitempty"`
+	// ConfigDir is the directory storing the file-based catalog contents.
+	ConfigDir string `json:"configDir,omitempty	"`
```
joelanford, line 172: This `Config` terminology comes from the original `Declarative Configuration` name we gave to this stuff, which was later renamed to `File-based catalogs`.

Should we call this `CatalogDir`?

PR #298, opened 2023-09-28
Add config for packageserver wakeup interval

tmshort:


pkg/operators/v1/olmconfig_test.go
```diff
@@ -2,14 +2,85 @@ package v1
...
 
 func boolPointer(in bool) *bool {
 	return &in
 }
+func TestPackageServerWakeupInterval(t *testing.T) {
+	five := time.Minute * 5
+	one := time.Second * 60
 
+	fiveParsed, err := time.ParseDuration("5m")
+	assert.NoError(t, err)
```
ncdc, line 20:
```suggestion
	require.NoError(t, err)
```

pkg/operators/v1/olmconfig_test.go
```diff
@@ -2,14 +2,85 @@ package v1
...
+
+	oneParsed, err := time.ParseDuration("60s")
+	assert.NoError(t, err)
```
ncdc, line 23:
```suggestion
	require.NoError(t, err)
```

pkg/operators/v1/olmconfig_types.go
```diff
@@ -23,6 +25,9 @@ type Features struct {
 	// When reenabled, OLM will recreate the "Copied CSVs" for each
 	// cluster scoped operator.
 	DisableCopiedCSVs *bool `json:"disableCopiedCSVs,omitempty"`
+	// PackageServerWakeupInterval is used to define the wakeup interval
+	// for packagerservers
```
ncdc, line 29: Would it be possible to make this more descriptive?

tmshort: Updated.

tmshort:
The one thing missing here is validation; it gets parsed as a duration, but doesn't validate beforehand.

ncdc:
Try `// +kubebuilder:validation:Format=duration`

crds/operators.coreos.com_olmconfigs.yaml
```diff
@@ -43,6 +43,9 @@ spec:
                     disableCopiedCSVs:
                       description: DisableCopiedCSVs is used to disable OLM's "Copied CSV" feature for operators installed at the cluster scope, where a cluster scoped operator is one that has been installed in an OperatorGroup that targets all namespaces. When reenabled, OLM will recreate the "Copied CSVs" for each cluster scoped operator.
                       type: boolean
+                    packageServerWakeupInterval:
```
awgreene, line 46: Would `packageServerPeriodicSyncInterval` be more appropriate?

dinhxuanvu:
While this config is on olmconfig api which users choose to do this, I still feel a bit concerned that if they set this duration too longer (either intentionally or accidentally), the catalog data in package server is gonna be stale for a long time. I wonder if there is a way to validate and warn against this. Having said that, this should only impact the consumer of packageserver which is console at the moment. This shouldn’t impact the runtime resolution which is more important.

tmshort:
> Try `// +kubebuilder:validation:Format=duration`
>
>meh: https://github.com/kubernetes/apiextensions-apiserver/issues/56

I found this, and reduced it (before @dinhxuanvu even said anything) to limit it to hours/minutes/seconds.
https://github.com/openshift/hive/pull/1684

crds/operators.coreos.com_olmconfigs.yaml
tmshort: It's called wakeupInterval in the code... but I really don't care.

tmshort: "Periodic" and "Interval" are redundant. SO, perhaps just "packageServerSyncInterval"?

dinhxuanvu:
> > Try `// +kubebuilder:validation:Format=duration`
> > meh: [kubernetes/apiextensions-apiserver#56](https://github.com/kubernetes/apiextensions-apiserver/issues/56)
> 
> I found this, and reduced it (before @dinhxuanvu even said anything) to limit it to hours/minutes/seconds. [openshift/hive#1684](https://github.com/openshift/hive/pull/1684)
I can live with hours restrictions. At least that will allow daily update of catalog content.

tmshort:
> > > Try `// +kubebuilder:validation:Format=duration`
> > > meh: [kubernetes/apiextensions-apiserver#56](https://github.com/kubernetes/apiextensions-apiserver/issues/56)
> > 
> > 
> > I found this, and reduced it (before @dinhxuanvu even said anything) to limit it to hours/minutes/seconds. [openshift/hive#1684](https://github.com/openshift/hive/pull/1684)
> 
> I can live with hours restrictions. At least that will allow daily update of catalog content.

There's no limit to the number _before_ the hours, so it could be e.g. `72h`

dinhxuanvu:
> > > > Try `// +kubebuilder:validation:Format=duration`
> > > > meh: [kubernetes/apiextensions-apiserver#56](https://github.com/kubernetes/apiextensions-apiserver/issues/56)
> > > 
> > > 
> > > I found this, and reduced it (before @dinhxuanvu even said anything) to limit it to hours/minutes/seconds. [openshift/hive#1684](https://github.com/openshift/hive/pull/1684)
> > 
> > 
> > I can live with hours restrictions. At least that will allow daily update of catalog content.
> 
> There's no limit to the number _before_ the hours, so it could be e.g. `72h`

I’m aware of that but it’s unlikely that they mistakenly add that big of a number though it could still be a typo. That at least reduces the risk of having days of waiting if we allow days or beyond. I would prefer to restrict it to mins only but then that may sound too restrictive. I think this is reasonable compromise.
We can do further by issuing warning if the number exceeds 24 hrs but that is up to you.

tmshort:
> I’m aware of that but it’s unlikely that they mistakenly add that big of a number though it could still be a typo. That at least reduces the risk of having days of waiting if we allow days or beyond. I would prefer to restrict it to mins only but then that may sound too restrictive. I think this is reasonable compromise. We can do further by issuing warning if the number exceeds 24 hrs but that is up to you.

It seemed telcos wanted to increase the period to days, and this is meant for IoT-type devices (which again, they may want to update reasonably quickly).

dinhxuanvu:
> > I’m aware of that but it’s unlikely that they mistakenly add that big of a number though it could still be a typo. That at least reduces the risk of having days of waiting if we allow days or beyond. I would prefer to restrict it to mins only but then that may sound too restrictive. I think this is reasonable compromise. We can do further by issuing warning if the number exceeds 24 hrs but that is up to you.
> 
> It seemed telcos wanted to increase the period to days, and this is meant for IoT-type devices (which again, they may want to update reasonably quickly).

Given this is for single-node/telco/LP usage, I wonder if their intention is to suppress the packageserver as much as possible because they simply don't use console at all. Deploying via CLI is usually more common then point-and-click console for these types of deployments. If that is the case then frankly I don't care as much about safeguarding this feature given packageserver has nothing to do with runtime catalog update.

tmshort:
ping for approval/lgtm?

dinhxuanvu (approved):
/approve

dinhxuanvu:
> ping for approval/lgtm?

I approved this but will a secondary person to lgtm.

pkg/operators/v1/olmconfig_types.go
```diff
@@ -23,6 +25,16 @@ type Features struct {
...
 	DisableCopiedCSVs *bool `json:"disableCopiedCSVs,omitempty"`
+	// PackageServerSyncInterval is used to define the sync interval for
+	// packagerserver pods. Packageserver pods periodically check the
+	// status of CatalogSources; this specifies the period using duration
+	// format (e.g. "60m"). For this parameter, only hours ("h"), minutes
+	// ("m"), and seconds ("s") may be specified. When not specified, the
+	// period defaults to the value specified within the packageserver.
+	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(s|m|h))+$"
```
ncdc, line 36: For this part of the regex, `(\\.[0-9]+)?`, it matches a literal `\`, then any character, then a digit; 0 or 1 time. This means that for the regex as a whole, `1\a3m` matches. I'm guessing this isn't intended?

tmshort: It doesn't match the literal `\`... the double slash is in there because it's a comment, and thus the escape for the `.` needs to be escaped itself.
I pulled it from https://github.com/openshift/hive/pull/1684 
see: https://github.com/openshift/hive/blob/227c25916da4c0b1bb3255b7c030d2ccf36384a0/apis/hive/v1/clusterdeployment_types.go#L187

crds/operators.coreos.com_olmconfigs.yaml
```diff
@@ -43,6 +43,10 @@ spec:
                     disableCopiedCSVs:
                       description: DisableCopiedCSVs is used to disable OLM's "Copied CSV" feature for operators installed at the cluster scope, where a cluster scoped operator is one that has been installed in an OperatorGroup that targets all namespaces. When reenabled, OLM will recreate the "Copied CSVs" for each cluster scoped operator.
                       type: boolean
+                    packageServerSyncInterval:
+                      description: PackageServerSyncInterval is used to define the sync interval for packagerserver pods. Packageserver pods periodically check the status of CatalogSources; this specifies the period using duration format (e.g. "60m"). For this parameter, only hours ("h"), minutes ("m"), and seconds ("s") may be specified. When not specified, the period defaults to the value specified within the packageserver.
+                      type: string
+                      pattern: ^([0-9]+(\.[0-9]+)?(s|m|h))+$
```
tmshort, line 49: @ncdc: here's the expression in the YAML.

dinhxuanvu:
> > I approved this but will need a secondary person to lgtm.
> 
> @dinhxuanvu why?

Usually we require two separate people to review and thumb up on the PR especially involving an API change at least that's how we worked back in my time. I'm still following that rule. That's why I didn't approve and lgtm at the same time. Plus, I wanted other people to have a chance to block this PR if they didn't agree with me.

pkg/operators/v1/olmconfig_types.go
stevekuznetsov: Don't ask how I ended up here ... but FYI @tmshort if you use back-tick quotes you don't need to do the escaping:

```
// +kubebuilder:validation:Pattern=`^([0-9]+(\.[0-9]+)?(s|m|h))+$`
```

PR #304, opened 2023-11-01
Separate Category and Capability Validation

kevinrizza:
This commit removes the default set of category and capability validation as a part of the operatorhubio validator.

As we have a mechanism for custom category validation, and there is significantly more churn on that specific validation, this commit separates the default operatorhubio validator from a distinct default categories validator. This allows users that want to continue to use the default set of categories to still do so, and if there are custom categories they would like to include they are free to use the dynamic categories validation option instead.

This commit also does the same separation for capability validation. There is no implementation of custom capability validation (as there is less churn and no explicit need for that yet) -- adding custom capability validation should be trivial in a future commit.

This commit accomplishes this by deprecating the existing validator and creating a v2 version of the operatorhubio validator.

Additionally, this commit adds 'Observability' to the category list.


pkg/validation/internal/operatorhub.go
```diff
@@ -221,7 +228,8 @@ func validateHubCSVSpec(csv v1alpha1.ClusterServiceVersion) CSVChecks {
 	checks = checkSpecProviderName(checks)
 	checks = checkSpecMaintainers(checks)
 	checks = checkSpecLinks(checks)
-	checks = checkAnnotations(checks)
+	checks = checkCapabilities(checks)
```
joelanford, line 231: Capabilities seems like another thing worth breaking out into its own validator. WDYT?

pkg/validation/internal/operatorhub.go
```diff
@@ -178,6 +180,11 @@ func validateOperatorHub(objs ...interface{}) (results []errors.ManifestResult)
 		}
 	}
 
+	// Add a deprecation warning to the list so that users are aware this validator is deprecated
+	deprecationResultWarning := errors.ManifestResult{}
+	deprecationResultWarning.Add(errors.WarnDeprecatedValidator("OperatorHub Validator deprecated, use OperatorHubV2 validator"))
```
joelanford, line 185: Two suggestions:
1. Use the name of the validator as it will show up for a user (see [here](https://github.com/operator-framework/operator-sdk/blob/d21ed6499ebfc8ecdb4508e1c2a2a0cfd2a151f3/internal/cmd/operator-sdk/bundle/validate/optional.go#L40); as an aside, it irks me that the name doesn't live with the code.. are these validators used elsewhere?)
2. Mention the combination of validators that are required to get equivalent validation

Also open to suggestions on how to format the `v2` suffix. Operator SDK/Kubebuilder use `<name>/<version>` for plugin names, so this is probably the most predictable for users.
```suggestion
	deprecationResultWarning.Add(errors.WarnDeprecatedValidator(`The "operatorhub" validator is deprecated; for equivalent validation use "operatorhub/v2" and "standardcategories" validators`))
```

pkg/validation/internal/operatorhub.go
```diff
@@ -118,7 +118,7 @@ import (
 // `k8s-version` key is allowed. If informed, it will perform the checks against this specific Kubernetes version where the
 // operator bundle is intend to be used and will raise errors instead of warnings.
 // Currently, this check is capable of verifying the removed APIs only for Kubernetes 1.22 version.
-var OperatorHubValidator interfaces.Validator = interfaces.ValidatorFunc(validateOperatorHub)
+var OperatorHubValidator interfaces.Validator = interfaces.ValidatorFunc(validateOperatorHubDeprecated)
```
joelanford, line 121: Add a GoDoc deprecation comment as well.
```suggestion
//
// Deprecated: Use OperatorHubV2Validator and StandardCategoriesValidator for equivalent validation.
var OperatorHubValidator interfaces.Validator = interfaces.ValidatorFunc(validateOperatorHubDeprecated)
```

pkg/validation/internal/operatorhubv2.go
```diff
@@ -0,0 +1,81 @@
+package internal
+
+import (
+	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/api/pkg/validation/errors"
+	interfaces "github.com/operator-framework/api/pkg/validation/interfaces"
+)
+
+var OperatorHubValidatorV2 interfaces.Validator = interfaces.ValidatorFunc(validateOperatorHubV2)
```
joelanford, line 10: Nit
```suggestion
var OperatorHubV2Validator interfaces.Validator = interfaces.ValidatorFunc(validateOperatorHubV2)
```

pkg/validation/validation.go
```diff
@@ -28,9 +28,17 @@ var CustomResourceDefinitionValidator = internal.CRDValidator
 var BundleValidator = internal.BundleValidator
 
 // OperatorHubValidator implements Validator to validate bundle objects
-// for OperatorHub.io requirements.
+// for OperatorHub.io requirements. This validator is deprecated.
 var OperatorHubValidator = internal.OperatorHubValidator
```
joelanford, line 32: If you format the deprecation message a bit different, I think IDEs and maybe even the GoDoc sites, pick up on it better:

```suggestion
// for OperatorHub.io requirements.
//
// Deprecated: Use OperatorHubV2Validator and StandardCategoriesValidator for equivalent functionality.
var OperatorHubValidator = internal.OperatorHubValidator
```

pkg/validation/validation.go
```diff
@@ -28,9 +28,17 @@ var CustomResourceDefinitionValidator = internal.CRDValidator
...
 
+// OperatorHubValidatorV2 implements Validator to validate bundle objects
+// for OperatorHub.io requirements.
+var OperatorHubValidatorV2 = internal.OperatorHubValidatorV2
```
joelanford, line 36: Nit: seems more consistent to put `Validator` as the suffix?
```suggestion
var OperatorHubV2Validator = internal.OperatorHubValidatorV2
```

pkg/validation/internal/operatorhub.go
```diff
@@ -151,13 +151,15 @@ var validCategories = map[string]struct{}{
 	"Security":                  {},
 	"Storage":                   {},
 	"Streaming & Messaging":     {},
+	"Observability":             {},
```
joelanford, line 154: Move this list to `standardcategories.go`?

pkg/validation/internal/operatorhub.go
```diff
@@ -267,6 +275,13 @@ func checkAnnotations(checks CSVChecks) CSVChecks {
 			checks.errs = append(checks.errs, fmt.Errorf("csv.Metadata.Annotations.Capabilities %s is not a valid capabilities level", capability))
 		}
 	}
+	return checks
+}
+
+func checkCategories(checks CSVChecks) CSVChecks {
```
joelanford, line 281: Move this function to `standardcategories.go`?

pkg/validation/internal/operatorhub.go
```diff
@@ -267,6 +275,13 @@ func checkAnnotations(checks CSVChecks) CSVChecks {
 			checks.errs = append(checks.errs, fmt.Errorf("csv.Metadata.Annotations.Capabilities %s is not a valid capabilities level", capability))
```
joelanford, line 275: Unrelated to this PR, but I noticed in the test you added that the word "Magic" was not quoted, which made it harder to deciper, for me at least:
```suggestion
			checks.errs = append(checks.errs, fmt.Errorf("csv.Metadata.Annotations.Capabilities %q is not a valid capabilities level", capability))
```

joelanford: Another similar nit for the categories message in lines 298 and 306.

Also in line 306, WDYT about `s/default/standard/`?

pkg/validation/internal/operatorhub.go
kevinrizza: I think this mapping is just a convention for the ux in the operatorsdk cli? The only thing this api exposes is the actual public interfaces, which can't have a slash. We'd have to generate these mappings as a wrapper type if we wanted to expose them in a generic way, I think. For now, I think just adding the message in this format seems reasonable though.

pkg/validation/validation.go
kevinrizza: nice catch

PR #307, opened 2023-11-10
Additional Deprecation Types

dtfranz:
Add additional deprecation conditions for each type (package, channel, bundle)

### DIFF
--- a/pkg/operators/v1alpha1/subscription_types.go
+++ b/pkg/operators/v1alpha1/subscription_types.go
@@ -118,8 +118,18 @@ const (
 	// SubscriptionBundleUnpackFailed indicates that the unpack job failed
 	SubscriptionBundleUnpackFailed SubscriptionConditionType = "BundleUnpackFailed"
 
-	// SubscriptionOperatorDeprecated indicates that the Operator currently installed with this Subscription has been deprecated.
-	SubscriptionOperatorDeprecated SubscriptionConditionType = "OperatorDeprecated"
+	// SubscriptionDeprecated is a roll-up condition which indicates that the Operator currently installed with this Subscription
+	//has been deprecated. It will be present when any of the three deprecation types (Package, Channel, Bundle) are present.
+	SubscriptionDeprecated SubscriptionConditionType = "Deprecated"
+
+	// SubscriptionOperatorDeprecated indicates that the Package currently installed with this Subscription has been deprecated.
+	SubscriptionPackageDeprecated SubscriptionConditionType = "PackageDeprecated"
+
+	// SubscriptionOperatorDeprecated indicates that the Channel used with this Subscription has been deprecated.
+	SubscriptionChannelDeprecated SubscriptionConditionType = "ChannelDeprecated"
+
+	// SubscriptionOperatorDeprecated indicates that the Bundle currently installed with this Subscription has been deprecated.
+	SubscriptionBundleDeprecated SubscriptionConditionType = "BundleDeprecated"
 )
 
 const (


pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -118,8 +118,18 @@ const (
 	// SubscriptionBundleUnpackFailed indicates that the unpack job failed
 	SubscriptionBundleUnpackFailed SubscriptionConditionType = "BundleUnpackFailed"
 
-	// SubscriptionOperatorDeprecated indicates that the Operator currently installed with this Subscription has been deprecated.
+	// SubscriptionOperatorDeprecated is a roll-up condition which indicates that the Operator currently installed with this Subscription
+	//has been deprecated. It will be present when any of the three deprecation types (Package, Channel, Bundle) are present.
 	SubscriptionOperatorDeprecated SubscriptionConditionType = "OperatorDeprecated"
```
joelanford, line 123: WDYT about renaming this one to just `Deprecated`?

I'm concerned that `OperatorDeprecated` sounds too much like `BundleDeprecated` and might confuse users.

dtfranz: Agreed, I've updated to `Deprecated`

pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -118,8 +118,18 @@ const (
...
-	// SubscriptionOperatorDeprecated indicates that the Operator currently installed with this Subscription has been deprecated.
-	SubscriptionOperatorDeprecated SubscriptionConditionType = "OperatorDeprecated"
+	// SubscriptionDeprecated is a roll-up condition which indicates that the Operator currently installed with this Subscription
+	//has been deprecated. It will be present when any of the three deprecation types (Package, Channel, Bundle) are present.
+	SubscriptionDeprecated SubscriptionConditionType = "Deprecated"
+
+	// SubscriptionOperatorDeprecated indicates that the Package currently installed with this Subscription has been deprecated.
+	SubscriptionPackageDeprecated SubscriptionConditionType = "PackageDeprecated"
+
+	// SubscriptionOperatorDeprecated indicates that the Channel currently installed with this Subscription has been deprecated.
```
everettraven, line 128: Nit: Channels aren't installable IIUC. Not serious enough to block this IMO.
```suggestion
	// SubscriptionOperatorDeprecated indicates that the Channel used with this Subscription has been deprecated.
```

everettraven (approved):

pkg/operators/v1alpha1/subscription_types.go
dtfranz: Woops, I'll correct this in a follow up. Thanks for catching.

pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -118,8 +118,18 @@ const (
 	// SubscriptionBundleUnpackFailed indicates that the unpack job failed
 	SubscriptionBundleUnpackFailed SubscriptionConditionType = "BundleUnpackFailed"
 
-	// SubscriptionOperatorDeprecated indicates that the Operator currently installed with this Subscription has been deprecated.
-	SubscriptionOperatorDeprecated SubscriptionConditionType = "OperatorDeprecated"
+	// SubscriptionDeprecated is a roll-up condition which indicates that the Operator currently installed with this Subscription
+	//has been deprecated. It will be present when any of the three deprecation types (Package, Channel, Bundle) are present.
+	SubscriptionDeprecated SubscriptionConditionType = "Deprecated"
```
dinhxuanvu, line 123: I’m not seeing the point of this condition. There are three distinct conditions that reflect which resource is deprecated. I don’t see the need of a blanket condition unless there is a usage for it that is not described in the comment or commit. This is why we should start to do proof PR to show how these fields are being used as a part of reviewing process.

dinhxuanvu: Or is this supposed to replace OperatorDeprecated?

pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -118,8 +118,18 @@ const (
 	// SubscriptionBundleUnpackFailed indicates that the unpack job failed
 	SubscriptionBundleUnpackFailed SubscriptionConditionType = "BundleUnpackFailed"
 
-	// SubscriptionOperatorDeprecated indicates that the Operator currently installed with this Subscription has been deprecated.
```
dinhxuanvu, line 121: I suspect this condition is used on olm side. So this is kind of an api change. Is it necessary to remove this without deprecation warning?

pkg/operators/v1alpha1/subscription_types.go
dtfranz: Yes, it's replacing `OperatorDeprecated` as the roll-up condition name. Having the roll-up allows us and any users to determine deprecated status without having to check for all three different sup-types.

pkg/operators/v1alpha1/subscription_types.go
dtfranz: For context, we just added that field yesterday and it's not in any release, so it's not in use anywhere. We're making a correction to the name here and also adding sub-types so that users can see which levels their installed Operators are deprecated at.

PR #308, opened 2023-11-17
Replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`

Juneezee:
The `github.com/ghodss/yaml` package is no longer actively maintained. There are numerous inquiries about the project's status on its issue tracker: https://github.com/ghodss/yaml/issues. `sigs.k8s.io/yaml` is a permanent fork of `github.com/ghodss/yaml`, which is actively maintained.

`sigs.k8s.io/yaml` was already an indirect dependency before this pull request was made, and `github.com/ghodss/yaml` is only used in tests files. We can safely replace it and remove 1 extra dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.19
 
 require (
 	github.com/blang/semver/v4 v4.0.0
-	github.com/ghodss/yaml v1.0.0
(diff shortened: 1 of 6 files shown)


joelanford (approved):
Thank you!

/lgtm
/approve

PR #312, opened 2024-01-19
Add field Subscription.Spec.Config.Annotations

cognifloyd:
I need a way to annotate operator deployments to satisfy an admissions hook in my on-prem vanilla (not OCP) k8s clusters (The admissions hook requires annotations that say which team is responsible for that deployment).

After discussing in [#olm-dev](https://kubernetes.slack.com/archives/C0181L6JYQ2), it looks like `Subscription` is the best user (cluster admin) facing API to extend for this purpose, as other resources, like CSV, should generally be opaque to the cluster admin.

So, this PR adds an `Annotations` field to `SubscriptionConfig` (`Subscription.Spec.Config.Annotations`). Is there a better place for this? Or a better name?

For reference, here is a draft matching change I plan to make in the operator-lifecycle-manager repo once this PR is merged and a new api release is available:
- https://github.com/operator-framework/operator-lifecycle-manager/compare/master...cognifloyd:operator-lifecycle-manager:deployment_annotations


pkg/operators/v1alpha1/subscription_types.go
```diff
@@ -90,6 +90,11 @@ type SubscriptionConfig struct {
 	// Use empty object ({}) to erase original sub-attribute values.
 	// +optional
 	Affinity *corev1.Affinity `json:"affinity,omitempty" protobuf:"bytes,18,opt,name=affinity"`
+
+	// DeploymentAnnotations is an unstructured key value map stored with a resource that may be
+	// set by external tools to store and retrieve arbitrary metadata.
+	// +optional
+	DeploymentAnnotations map[string]string `json:"deployment_annotations,omitempty" protobuf:"bytes,12,rep,name=deployment_annotations"`
```
joelanford, line 97: Kubernetes convention is to use camel case for field names.

```suggestion
	DeploymentAnnotations map[string]string `json:"deploymentAnnotations,omitempty" protobuf:"bytes,12,rep,name=deployment_annotations"`
```

crds/operators.coreos.com_subscriptions.yaml
```diff
@@ -524,6 +524,11 @@ spec:
                                   topologyKey:
                                     description: This pod should be co-located (affinity) or not co-located (anti-affinity) with the pods matching the labelSelector in the specified namespaces, where co-located is defined as running on a node whose value of the label with key topologyKey matches that of any node on which any of the selected pods is running. Empty topologyKey is not allowed.
                                     type: string
+                    deployment_annotations:
+                      description: DeploymentAnnotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata.
```
kevinrizza, line 528: Can we just call this annotations? I think it might be more useful to propogate annotations to all the resources that get created from the installation (rbac, apiservices, etc) than just deployments? Wdyt?

cognifloyd: That would be a much more invasive change. It would touch code that creates/updates:
- Deployment
- Pod (via the Deployment spec)
- Namespace
- Service
- CustomResourceDefinition
- APIService
- Secret
- ConfigMap
- ServiceAccount
- Role
- RoleBinding
- ClusterRole
- ClusterRoleBinding
- MutatingWebhookConfiguration
- ValidatingWebhookConfiguration
- custom resources: PrometheusRule, ServiceMonitor, PodDisruptionBudget, PriorityClass, ...

I'm not sure if my searches through the code of controllers missed any other resources. I only need to annotate deployments. If we call this `annotations`, would you accept a PR where I implement only the deployment piece, leaving the rest for someone else to contribute later? Or would the PR have to touch logic around all these other resources before it could be merged?

Thinking about UX: everything else in SubscriptionConfig explicitly configures the Pod spec in Deployments, not any other resource. Would calling this `annotations` suggest that this too will only apply to the pods?

cognifloyd: @kevinrizza If you do want one `annotations` field for all resources, then what about just propagating the `annotations` field of `Subscription` instead of adding a new field to `SubscriptionConfig`?

kevinrizza: You're getting at why I think this change is actually a hard one for us to maintain :)

My preference would be to rename this to just `annotations` and have the implementation just apply to `deployments`, `apiservices` and `pods`. The reconciliation of those all happens in more or less the same place. If someone else comes along and needs these annotations to apply to their rbac then we can deal with that when we get there.

cognifloyd: Renamed.

kevinrizza (approved):

PR #324, opened 2024-05-22
:bug: fix ClusterServiceVersion CRD Generation

perdasilva:
New API updates is tripping up more validations against default values in the Deployment stanza of the CSV. This PR adds a couple of yq replacement queries to add default values to those fields.

seems the crds/zz_defs.go was also out of sync - regeneration seems to have fixed it


perdasilva:
I think we can override the verify here. It seems like there's been a change in the go doc for json.RawMessage:

```
// RawMessage is a raw encoded JSON value.
// It implements [Marshaler] and [Unmarshaler] and can
// be used to delay JSON decoding or precompute a JSON encoding.
type RawMessage []byte
```
https://github.com/golang/go/blob/release-branch.go1.22/src/encoding/json/stream.go#L255

This is changing the field description and causing verify to fail

perdasilva:
> It seems like `crds/zz_defs.go` needs `make format`

Thank you! I was chasing a different issue. Let's see if verify clears it (it will fail for the reasons I stated above in olm)...

m1kola (approved):

m1kola:
/hold

One to check one thing before we merge

PR #325, opened 2024-05-22
:seedling: Add GHA to test CRDs against kind

perdasilva:
Currently, the api repo is green but the CRDs cannot be applied. This PR adds a job to the repo to check that the generated CRDs can be applied to a kind cluster running on the same kube version as the client-go in go.mod

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,10 +101,18 @@ $(LOCALBIN):
 ## Tool Binaries
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 YQ ?= $(LOCALBIN)/yq
+KIND ?= $(LOCALBIN)/kind
 
 ## Tool Versions
 CONTROLLER_TOOLS_VERSION ?= v0.15.0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -115,3 +123,13 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 yq: $(YQ) ## Download yq locally if necessary.
 $(YQ): $(LOCALBIN)
 	GOBIN=$(LOCALBIN) go install $(GO_INSTALL_OPTS) github.com/mikefarah/yq/v4@$(YQ_VERSION)
+
+.PHONY: kind
+kind: $(KIND) ## Download yq locally if necessary.
+$(KIND): $(LOCALBIN)
+	GOBIN=$(LOCALBIN) go install $(GO_INSTALL_OPTS) sigs.k8s.io/kind@latest
```
m1kola, line 130: Nit: Might be worth adding bingo into this repo.

m1kola (approved):

PR #337, opened 2024-05-22
Perdasilva/bump cel 2

perdasilva:


perdasilva:
cannot bump cel version as it must follow the same version as the apiextension-server

PR #340, opened 2024-06-07
bugfix for catalog-operator and packageserver pods being in CrashLoopBackOff if updateStrategy is empty

rashmigottipati:
Description: 

- dont poll if updateStrategy and any of its child references are empty
- cover the panic by adding nil protection checks along the pointer stack of updateStrategy struct


perdasilva:
closing as this is no longer required

PR #342, opened 2024-06-13
remove default value of catsrc.spec.grpcPodConfig.securityContextConfig

joelanford:
In https://github.com/operator-framework/operator-lifecycle-manager/pull/3206, we updated OLM to support an implicit default of a catalog pod's securityContextConfig, based on the PSA configuration of the namespace it is created in.

However, we overlooked the fact that the CatalogSource CRD specifies `legacy` as a default value. The result of this is that _any_ catalog source with a non-nil `spec.grpcPodConfig` that attempts to leave `spec.grpcPodConfig.securityContextConfig` unset will be defaulted to legacy during admission.

This PR fixes that by removing the default value in the CRD spec.


grokspawn (approved):

acornett21 (approved):

crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -1023,19 +1023,15 @@ spec:
...
+                        run in PSA `baseline` or `privileged` namespaces. If the SecurityContextConfig is unspecified, the mode will be
+                        determined by the namespace's PSA configuration. If the namespace is enforcing `restricted` mode, then the pod
+                        will be configured as if `restricted` was specified. Otherwise, it will be configured as if `legacy` was
+                        specified. Specifying a value other than `legacy` or `restricted` result in a validation error. When using older
+                        catalog images, which can not run in `restricted` mode, the SecurityContextConfig should be set to `legacy`.
 
 
                         More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'
                       type: string
-                      default: legacy
```
jianzhangbjz, line 1038: Hi, one more question, why not set the default value to `restricted`? Thanks!

jianzhangbjz: Got it, thanks!

PR #350, opened 2024-07-02
CSV: Add annotations into StrategyDeploymentSpec

hongkailiu:
The intension is to add the annotations to the deployment ALM creates, like the labels.


kevinrizza:
@hongkailiu Thanks for the contribution! Can you tell me the reason for this contribution? We're actively trying to move OLM v0 (which includes the CSV API) into a feature complete state, so we're trying to limit what gets added to the APIs at this point. What are you trying to solve for by adding hardcoded annotations to CSVs as opposed to using the subscription config at runtime on a cluster?

kevinrizza:
@hongkailiu Following along here https://issues.redhat.com/browse/OTA-1166 I think there's an even simpler option. OLM will propogate annotations defined on the CSV meta itself to the deployments it creates (see https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/deployment.go#L147).

So you could just annotate the CSV instead of requiring an update to the API here. Let me know if that's sufficient, and if so I think this pull request isn't needed. We appreciate the work you put into it though!

hongkailiu:
Thanks @kevinrizza for the review.

My intension is this:
https://github.com/operator-framework/operator-lifecycle-manager/pull/3339

And then:
I expect that the annotations from the manager of this controller
https://github.com/openshift/cincinnati-operator/pull/185/files#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480R5
will appear in the CSV when it is generated by [this command](https://github.com/openshift/cincinnati-operator/blob/e44857a46e222cd30eff7a43c1ffa652335d7d11/Makefile#L106):

> $(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)

The goal of all this above is that when a deployment is created via the CSV, it has the annotation that we want.

I believe [this](https://github.com/operator-framework/operator-lifecycle-manager/pull/2054) is a similar PR for labels.
The current PR is for the annotations.

I am new to the operator world.
Feedback is appreciated if I did not do it correctly to achieve my goal above.

hongkailiu:
> So you could just annotate the CSV instead of requiring an update to the API here. Let me know if that's sufficient, and if so I think this pull request isn't needed. We appreciate the work you put into it though!

I also found this while reading the code.
Let me try this and get back to you.
If it works out, i would consider it good enough for me.

hongkailiu:
It works. Thanks for the workaround. @kevinrizza 

/close

PR #365, opened 2024-10-01
fix: add missing comment tags for Conditions

manusa:
## Descriptions

The tags for the Conditions field in BundleLookup are missing.
This causes issues while trying to generate OpenAPI specs for these structs.


jmrodri (approved):
/lgtm

PR #421, opened 2025-04-11
make grpcPodConfig.extractContent.cacheDir an optional field

grokspawn:


crds/operators.coreos.com_catalogsources.yaml
```diff
@@ -975,11 +975,10 @@ spec:
                         configured to use *must* be using the file-based catalogs in order to utilize this feature.
                       type: object
                       required:
-                        - cacheDir
                         - catalogDir
                       properties:
                         cacheDir:
-                          description: CacheDir is the directory storing the pre-calculated API cache.
+                          description: CacheDir is the (optional) directory storing the pre-calculated API cache.
```
camilamacedo86, line 981: Won’t we need to update all the places that use it to ensure the value is set when it isn’t provided?

grokspawn: The oddity of the o-f/api repo is that its testing is really in other repos.  Testing for this is in https://github.com/operator-framework/operator-lifecycle-manager/pull/3556

grokspawn:
Contributes to  operator-framework/operator-lifecycle-manager/issues/3257

acornett21 (approved):

PR #439, opened 2025-07-22
bump go version to fix https://pkg.go.dev/vuln/GO-2025-3751

grokspawn:


grokspawn:
Umm.... I meant to apply the override label, but did NOT mean to drop the lgtm/approve labels.  
I'll put them back.  
Sorry for the confusion!!

PR #493, opened 2026-05-26
fix(pkg/validation/internal): check nil pointers before dereference

alrs:
Several of the functions in `pkg/validation/internal` properly check to make sure that they haven't received a nil pointer, but they imporperly do it only after first attempting to dereference the pointer.


perdasilva:
@grokspawn it would be good if you could :eyes: this. I don't know if there was a reason to set the name in the result.

tmshort (approved):
/lgtm

PR #495, opened 2026-05-29
pkg/manifests: fix dropped walk errors

alrs:
A `filepath.Walk()` function takes an error as one of its arguments. Usually the first thing to do in a `Walk()` is to check if the `err` provided is not `nil` and should be returned. This fixes three places in `pkg/manifests` where a `WalkFunc()` was not handling the `err`.

Tests continue to pass.

### DIFF
--- a/pkg/manifests/bundleloader.go
+++ b/pkg/manifests/bundleloader.go
@@ -122,6 +122,9 @@ func collectWalkErrs(walk filepath.WalkFunc, errs *[]error) filepath.WalkFunc {
 }
 
 func (b *bundleLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err error) error {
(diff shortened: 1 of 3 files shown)


perdasilva:
@alrs Thanks for this ^^ the functions changed are exported - should we add a unit test for each for the changes?

tmshort (approved):
/lgtm

Looks fine to me, these types of walk functions should be returning the error they receive as input.

tmshort:
/approve cancel
/hold cancel

I did not mean to LGTM _and_ approve

perdasilva:
@tmshort @alrs I've added a commit with unit tests - let me know if this is an improvement or not

rashmigottipati (approved):
/lgtm

PR #524, opened 2026-09-15
[wip]OPRUN-4764: CSV: Extend RelatedImage with labels

hongkailiu:
Add the optional `labels` field to the ClusterServiceVersion type RelatedImage, so operator authors can classify related images by product feature in the bundle's CSV.

The field is optional and additive: CSVs that do not use it are unchanged, and the meaning of the existing `name` and `image` fields is unchanged. Keys and values follow the Kubernetes label syntax and constraints; the syntax is validated by `opm validate`, not here.

Jira: OPRUN-4764

## Summary by CodeRabbit

* **New Features**
  * Related images can now include Kubernetes-style labels for organization and selection.
  * Label metadata is preserved when related-image data is copied or converted between formats.
  * Mirroring tools can use label selectors to choose labeled related images; unlabeled images continue to be mirrored automatically.
  * Bundle authors can restore related-image labels after generation when their workflow requires them.


grokspawn:
I think you might also need to provide syntax validation in the validators this project supplies (for e.g. https://github.com/operator-framework/api/blob/master/pkg/validation/internal/bundle.go).

